PR #248, opened 2026-03-04
fix: add preflight guardrail for suspicious amount scaling (closes #219)

### LINKED ISSUE #219: Add preflight guardrail for suspicious amount scaling (unit mismatch)
## Problem
A unit mismatch can produce interpreted amounts that are far above wallet balance. This currently fails late and can waste agent retries/thinking.

## Proposal
Add a preflight guardrail before tx construction:
- If interpreted amount is suspiciously above available balance, fail fast with a clear unit-mismatch hint.
- Provide deterministic error code and remediation examples.

Suggested behavior:
- Error code: `AMOUNT_UNIT_MISMATCH_SUSPECTED`
- Message includes:
  - requested input
  - interpreted human/base values
  - wallet balance
  - corrected example for both units

## Acceptance Criteria
- [ ] Guard runs for quote+swap preflight where balance is available
- [ ] Deterministic error code emitted
- [ ] Error text includes corrected examples
- [ ] Threshold/config documented (or sensible hardcoded rule)

JackBinswitch-btc:
## Problem

When an agent calls `bitflow_get_quote` or `bitflow_swap` with `amountUnit="human"` (the default) but accidentally passes a base-unit value — e.g. `amountIn="1000000"` intending 1 STX but actually requesting 1,000,000 STX — the SDK receives an inflated amount and the transaction fails on-chain after burning STX fees and retries. The error is cryptic and gives no guidance on how to fix the call.

## Solution

Add a `checkAmountScaling()` preflight function, called immediately after `resolveAmountIn()` in both the `bitflow_get_quote` and `bitflow_swap` handlers.

**Logic:**
1. Fetch the wallet's current token balance via the Hiro API (STX via `getStxBalance`, FTs via `getTokenBalance`).
2. If `humanAmount > 10 × walletBalanceHuman`, throw `AMOUNT_UNIT_MISMATCH_SUSPECTED`.
3. The error message includes:
   - The requested amount and wallet balance
   - Token symbol and decimals
   - Two corrected call examples: one with `amountUnit="base"`, one with the properly scaled `amountUnit="human"` value

**Graceful degradation:** The check is silently skipped when:
- No wallet is unlocked / no mnemonic is set (cannot resolve address)
- The Hiro balance API is unreachable (network error)

This means a transient API outage never blocks legitimate swaps.

## Changes

Only `src/tools/bitflow.tools.ts` is modified:
- Added `import { getHiroApi }` from the existing hiro-api service
- Added `AMOUNT_SCALE_SUSPICION_MULTIPLIER = 10` constant
- Added `checkAmountScaling()` function (~80 lines, fully documented)
- Added one `await checkAmountScaling(...)` call in `bitflow_get_quote`
- Added one `await checkAmountScaling(...)` call in `bitflow_swap`

TypeScript compiles cleanly (`tsc --noEmit`).

## Test plan

- [ ] Call `bitflow_get_quote` with `amountIn="1000000"` and `amountUnit="human"` on a wallet with ~1 STX balance — expect `AMOUNT_UNIT_MISMATCH_SUSPECTED` error with corrected examples
- [ ] Call the same with `amountUnit="base"` and `amountIn="1000000"` — expect normal quote for 1 STX (no guard fires because unit is explicit)
- [ ] Call with `amountIn="1"` and `amountUnit="human"` on a wallet with ~1 STX balance — expect normal quote (amount ≤ threshold)
- [ ] Call `bitflow_swap` with a suspicious amount — expect the same early rejection before any on-chain tx is constructed
- [ ] Simulate wallet-locked state (no mnemonic set) with a suspicious amount — expect normal pass-through (guard skipped)

Closes #219

### DIFF
--- a/src/tools/bitflow.tools.ts
+++ b/src/tools/bitflow.tools.ts
@@ -15,6 +15,33 @@ const HIGH_IMPACT_THRESHOLD = 0.05; // 5%
 const AMOUNT_SCALE_SUSPICION_MULTIPLIER = 10;
 
 /**
+ * Thrown by checkAmountScaling when the requested amount looks like it was
+ * supplied in base units instead of human units.
+ */
+class AmountUnitMismatchError extends Error {
+  readonly code = "AMOUNT_UNIT_MISMATCH_SUSPECTED";
+  readonly details: {
+    requestedAmountHuman: number;
+    requestedAmountInput: string;
+    walletBalanceHuman: number;
+    tokenDecimals: number;
+    tokenSymbol: string;
+    correctedHumanAmount: number;
+    correctedBaseAmount: string;
+    suspicionMultiplier: number;
+  };
+
+  constructor(
+    message: string,
+    details: AmountUnitMismatchError["details"]
+  ) {
+    super(message);
+    this.name = "AmountUnitMismatchError";
+    this.details = details;
+  }
+}
+
+/**
  * Resolve amountIn to the human-unit number the Bitflow SDK expects.
  * When amountUnit is "human" (default), validates and passes through.
  * When amountUnit is "base", converts from smallest units using token decimals.
@@ -89,18 +116,19 @@ async function checkAmountScaling(
     return;
   }
 
-  // Look up the token's decimals from the Bitflow token list.
-  // Fall back to 6 (micro-STX standard) if the token is unknown.
-  const tokens = await bitflowService.getAvailableTokens();
-  const tokenMeta = tokens.find((t) => t.id === tokenX);
-  const decimals = tokenMeta?.decimals ?? 6;
-
-  // Fetch the wallet's balance for this token.
-  // STX is identified in Bitflow as "token-stx".
-  const hiroApi = getHiroApi(NETWORK);
-  let walletBalanceHuman: number;
-
   try {
+    // Look up the token's decimals from the Bitflow token list.
+    // Fall back to 6 (micro-STX standard) if the token is unknown.
+    // NOTE: getAvailableTokens() is inside this try/catch so that a Bitflow
+    // API outage does not propagate as an uncaught error and block all swaps.
+    const tokens = await bitflowService.getAvailableTokens();
+    const tokenMeta = tokens.find((t) => t.id === tokenX);
+    const decimals = tokenMeta?.decimals ?? 6;
+
+    // Fetch the wallet's balance for this token.
+    // STX is identified in Bitflow as "token-stx".
+    const hiroApi = getHiroApi(NETWORK);
+    let walletBalanceHuman: number;
     const isStx = tokenX === "token-stx";
     if (isStx) {
       const stxInfo = await hiroApi.getStxBalance(walletAddress);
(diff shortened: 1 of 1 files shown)


arc0btc (changes requested):
## Review: fix: add preflight guardrail for suspicious amount scaling

The intent is solid — early rejection with corrected examples is exactly what agents need to avoid fee-burning retries on unit mismatches. The `resolveAmountIn` + `checkAmountScaling` layering is clean, and the graceful-skip logic (no wallet, `amountUnit="base"`) is well-reasoned.

One correctness issue blocks the merge:

---

### 🔴 `getAvailableTokens()` is outside the try/catch — breaks the graceful degradation promise

The PR description says:

> "The check is silently skipped when... The balance fetch itself fails — we treat network errors as non-fatal so that a temporary API outage does not block all swap traffic."

But the token list fetch is **not** inside the try/catch:

```typescript
// NOT protected — throws propagate and block the swap
const tokens = await bitflowService.getAvailableTokens();
const tokenMeta = tokens.find((t) => t.id === tokenX);
const decimals = tokenMeta?.decimals ?? 6;

// Only this is protected
try {
  const isStx = tokenX === "token-stx";
  ...walletBalanceHuman = ...
} catch {
  return; // graceful
}
```

If `getAvailableTokens()` throws (Bitflow API down, SDK cache miss on cold start), `checkAmountScaling` throws uncaught, and **all** `bitflow_get_quote` / `bitflow_swap` calls fail. This is the opposite of graceful degradation — an API outage in the guardrail becomes a full swap outage.

Fix: wrap the entire function body (after the wallet-address check) in a try/catch that returns on any error, OR scope the token list fetch into the existing try/catch. The decimal fallback `?? 6` already handles an unknown token — it just doesn't handle a thrown error.

```typescript
try {
  const tokens = await bitflowService.getAvailableTokens();
  const tokenMeta = tokens.find((t) => t.id === tokenX);
  const decimals = tokenMeta?.decimals ?? 6;

  const hiroApi = getHiroApi(NETWORK);
  // ...balance fetch...
  // ...threshold check...
} catch {
  // Any network error in the guardrail is non-fatal — skip the check
  return;
}
```

---

### 🟡 Minor: `(error as any)` — use a typed error class

```typescript
(error as any).code = "AMOUNT_UNIT_MISMATCH_SUSPECTED";
(error as any).details = { ... };
```

With strict TS, a custom error class keeps types clean and makes the code easier to catch upstream:

```typescript
class AmountUnitMismatchError extends Error {
  code = "AMOUNT_UNIT_MISMATCH_SUSPECTED";
  details: { ... };
  constructor(message: string, details: ...) { ... }
}
```

Not blocking on its own, but the `any` cast is a smell in a strict-mode file.

---

### ✅ What's correct

- Guard only fires for `amountUnit="human"` — explicit base-unit callers bypass cleanly
- Error message includes three corrected call examples with real wallet balance context — exactly the right UX for agents that need deterministic recovery hints
- Called after `resolveAmountIn`, so `humanAmount` is always the normalized value — no double-conversion risk
- Zero-balance wallet skip is correct (no denominator to compare against)
- Issue #219 acceptance criteria are all met in the happy path

---

Fix the `getAvailableTokens()` try/catch scope and this is good to merge.

PR #265, opened 2026-03-04
feat: fix inscription fees and add parent/child support

### LINKED ISSUE #253: Feature Request: Parent/Child Inscription Support (Ordinals Provenance)
cc @biwas @whoabuddy @arc0btc

## Context

[AIBTC News](https://aibtc.news) inscribes daily intelligence briefs to Bitcoin using the MCP `inscribe` / `inscribe_reveal` tools. We want to use [Ordinals provenance](https://docs.ordinals.com/inscriptions/provenance.html) (parent/child inscriptions) to link every daily brief as a child of a single parent inscription, creating a verifiable on-chain collection.

Casey Rodarmor (Ordinals creator) specifically recommended this pattern for our use case.

**Parent inscription (already created):** [5f82a985e3797d3e6638443d6acc93ff4f27e574402bbf204966f2c54b4941b8i0](https://ordinals.com/inscription/5f82a985e3797d3e6638443d6acc93ff4f27e574402bbf204966f2c54b4941b8i0)

```
AIBTC.news
Daily agent news, inscribed permanently to Bitcoin.
```

## What's needed

The [provenance spec](https://docs.ordinals.com/inscriptions/provenance.html) requires three changes to the reveal transaction:

1. **Tag `3` in the inscription envelope** — serialized parent inscription ID (32-byte reversed txid + LE index, trailing zeroes stripped)
2. **Parent UTXO spent as an additional input** — proves ownership of the parent
3. **Parent returned as first output** — so it can be reused for the next child

## Proposed parameter additions

**`inscribe` (commit step):**
```
+ parentInscriptionId: string (optional) — e.g. "5f82a985...i0"
```

**`inscribe_reveal` (reveal step):**
```
+ parentInscriptionId: string (optional) — encoded as tag 3 in envelope
+ parentUtxo: string (optional) — "txid:vout" of parent inscription
+ parentUtxoValue: number (optional) — satoshi value (needed for signing)
```

**`estimate_inscription_fee`:**
```
+ parentInscriptionId: string (optional) — adds ~65-70 vbytes
```

## Reveal transaction structure

```
INPUTS:
  [0]  Parent inscription UTXO    ← new (key-path Schnorr signature)
  [1]  Commit output UTXO         ← existing (script-path spend)

OUTPUTS:
  [0]  Parent return → wallet     ← new (preserves sat position)
  [1]  Child inscription → wallet ← existing
```

## Existing JS/TS library

[`micro-ordinals`](https://github.com/paulmillr/micro-ordinals) (by paulmillr) already implements parent/child envelope construction in TypeScript, including tag `3` encoding and the `InscriptionId` serialization format. Uses `@scure/btc-signer` for transaction building. Could be used directly or as reference:

```typescript
import { p2tr_ord_reveal, Inscription } from 'micro-ordinals';

const inscription: Inscription = {
  tags: {
    contentType: 'text/plain;charset=utf-8',
    parent: '5f82a985e3797d3e6638443d6acc93ff4f27e574402bbf204966f2c54b4941b8i0',
  },
  body: new TextEncoder().encode('child content'),
};

const revealPayment = p2tr_ord_reveal(pubkey, [inscription]);
```

## Reference implementations

- **ord (canonical, Rust):** [tag.rs](https://github.com/ordinals/ord/blob/master/src/inscriptions/tag.rs), [inscription.rs](https://github.com/ordinals/ord/blob/master/src/inscriptions/inscription.rs), [plan.rs](https://github.com/ordinals/ord/blob/master/src/wallet/batch/plan.rs)
- **micro-ordinals (JS/TS):** [github.com/paulmillr/micro-ordinals](https://github.com/paulmillr/micro-ordinals)
- **Mainnet example:** [OnChainMonkey Genesis](https://ordinals.com/inscription/fb162a46943e5d7d31d72ee2c8c850e66c1ca5d0d453068aa63883528285ed21i0) — parent with 10k children

## Technical notes

- All new params are optional — fully backward compatible
- Commit input index shifts from `0` to `1` when parent is present (affects sighash calculation)
- Parent inputs use standard Schnorr key-path signing; commit input uses tapscript script-path (existing)
- `get_ordinal_utxos` already can locate the parent UTXO

## Use case

```
AIBTC News (parent inscription)
├── Brief 2026-03-01 (child)
├── Brief 2026-03-02 (child)
├── Brief 2026-03-03 (child)
└── ...daily, permanently
```

### LINKED ISSUE #263: inscribe_reveal fee calculation bug: fee = 120 × feeRate regardless of content size
## Bug Summary

The `inscribe_reveal` (and likely `inscribe_child_reveal`) tool calculates the reveal transaction fee as `120 × feeRate` regardless of the actual transaction size or inscription content size. This produces dramatically underpaying transactions that may never confirm.

## Expected Behavior

The reveal fee should be calculated based on the actual transaction virtual size (vsize) multiplied by the requested fee rate. For an 8.7 KB HTML inscription, the reveal tx is ~2,317 vB, so at 6 sat/vB the fee should be ~13,900 sats.

## Actual Behavior

The fee is always `120 × feeRate`:

| feeRate | Expected Fee | Actual Fee | Effective Rate | Result |
|---------|-------------|-----------|---------------|--------|
| 6 sat/vB | ~13,900 sats | 720 sats | 0.31 sat/vB | Stuck in mempool |
| 2 sat/vB | ~4,634 sats | 240 sats | 0.10 sat/vB | Stuck in mempool |

The remaining sats from the `revealAmount` go into the inscription output UTXO instead of being used as miner fee.

## Reproduction

1. Call `inscribe` with `contentType: "text/html;charset=utf-8"` and ~8.7 KB base64 content, `feeRate: 6`
2. Wait for commit to confirm
3. Call `inscribe_reveal` with the same parameters plus `commitTxid` and `revealAmount` from step 1
4. Check the broadcast reveal tx on mempool.space — the fee will be exactly `120 × feeRate` sats, and the effective sat/vB will be far below what was requested

## Impact

- **Two parent inscriptions stuck in mempool** at 0.10 and 0.31 sat/vB, likely to be evicted
- ~30,000 sats in commit outputs effectively lost (locked in taproot script-path outputs that only the reveal can spend)
- No way to CPFP via MCP tools because `psbt_sign` cannot sign taproot key-path inputs (missing tap internal key metadata)
- The `estimate_inscription_fee` tool returns the **correct** fee calculation, making the bug even more confusing — users see a correct estimate, then the actual tx underpays by ~27x

## Affected Transactions

- Reveal `26f5cce3a378d76e57a09cd0c7ca98a5eb10f1a4787eafa65f2bfcf16a9a92f4` — 240 sat fee, 0.10 sat/vB effective
- Reveal `912cbbf3c6b9e4a46ceec9c08514fee625b65e4b7469aacb73a54175b8fe6f39` — 720 sat fee, 0.31 sat/vB effective

## Suggested Fix

In the reveal transaction builder, the fee should be calculated as:
```
fee = revealTxVsize × feeRate
outputValue = 546  // dust limit for inscription UTXO
revealAmount = fee + outputValue
```

Currently it appears to use a hardcoded vsize estimate of 120 vB, which is only valid for tiny inscriptions (~50 bytes). The actual vsize scales with content size since the inscription data is in the witness.

## Environment

- aibtc-mcp-server (latest as of 2026-03-04)
- Mainnet
- Content: 8,668 bytes HTML inscription

### LINKED ISSUE #264: inscribe_child: parent inscription not found despite being confirmed on-chain
## Bug Description

`inscribe_child` returns `"Inscription {id} not found or missing UTXO data"` for a confirmed parent inscription that is visible on ordinals.com and has a confirmed UTXO at the expected taproot address.

## Steps to Reproduce

1. Create a parent inscription using `inscribe` + `inscribe_reveal` (not `inscribe_child`)
2. Wait for confirmation (in our case, confirmed at block 939,310)
3. Call `inscribe_child` with `parentInscriptionId: "912cbbf3c6b9e4a46ceec9c08514fee625b65e4b7469aacb73a54175b8fe6f39i0"`
4. Error: `Inscription 912cbbf3c6b9e4a46ceec9c08514fee625b65e4b7469aacb73a54175b8fe6f39i0 not found or missing UTXO data`

## Evidence the inscription exists

- **ordinals.com**: Returns HTTP 200 for `/inscription/912cbbf3c6b9e4a46ceec9c08514fee625b65e4b7469aacb73a54175b8fe6f39i0`
- **mempool.space**: UTXO confirmed at taproot address `bc1pcadvkp7l3mgtff2l9qk38e9wlapyfy830atrtq7dh6l3aqxa5snsczgzm7`, txid `912cbbf3...`, vout 0, value **18,226 sats**, block 939,310
- Multiple retries over several minutes — same error each time

## Possible Causes

1. **Non-standard UTXO value**: The parent inscription UTXO has 18,226 sats (due to the fee calculation bug in `inscribe_reveal` documented in #263). If the lookup filters by expected inscription output value (~546 sats), it would miss this UTXO.

2. **Internal database only**: The tool may only recognize inscriptions tracked in its own internal state, not arbitrary on-chain inscriptions. If the parent was created via `inscribe`/`inscribe_reveal` but the child lookup uses a different tracking mechanism, it could fail.

3. **Indexer lag**: The MCP server's ord indexer may not have synced to block 939,310 yet. However, the inscription has been confirmed for 20+ blocks, making this less likely.

## Related Issues

- #263 — `inscribe_reveal` fee calculation bug (same inscription affected)

## Impact

Cannot create child inscriptions linked to parent inscriptions created with this tool. Blocks the AIBTC News daily brief child inscription pipeline.

## Workaround

Building a standalone daemon that finds the parent UTXO directly via mempool.space API, bypassing the MCP tool's inscription lookup.

whoabuddy:
## Summary
- **Fix reveal fee calculation** — `buildRevealTransaction` was using the 34-byte P2TR output script size instead of actual inscription content size, causing all reveal fees to estimate at ~120 vbytes regardless of content (closes #263)
- **Add parent/child inscription support** — new builder logic for parent tag in envelope, parent UTXO as input 0, parent return as output 0, correct index shifting (closes #253)
- **Wire parent/child into MCP tools** — optional `parentInscriptionId` on `inscribe`, `inscribe_reveal`, and `estimate_inscription_fee` with Hiro Ordinals API lookup and validation (closes #264)
- **Refactor** — deduplicate fee rate resolution (`resolveBtcFeeRate` helper), extract `estimateRevealTxVbytes()`, remove redundant `hasParent` flag, fix broken tests

## Test plan
- [ ] `npm run build` compiles cleanly
- [ ] `npm test` — all 410 tests pass
- [ ] Test standard inscription flow (no parent) still works unchanged
- [ ] Test parent/child inscription with a known parent inscription ID
- [ ] Verify fee estimates scale with content size (not fixed ~120 vbytes)
- [ ] Verify parent not found / unconfirmed errors surface correctly

### DIFF
--- a/src/tools/ordinals.tools.ts
+++ b/src/tools/ordinals.tools.ts
@@ -20,8 +20,6 @@ import {
   P2TR_OUTPUT_VBYTES,
   TX_OVERHEAD_VBYTES,
   DUST_THRESHOLD,
-  P2TR_INPUT_BASE_VBYTES,
-  WITNESS_OVERHEAD_VBYTES,
 } from "../config/bitcoin-constants.js";
 import { createJsonResponse, createErrorResponse } from "../utils/index.js";
 import {
@@ -34,6 +32,7 @@ import {
   buildCommitTransaction,
   buildRevealTransaction,
   deriveRevealScript,
+  estimateRevealTxVbytes,
   type InscriptionData,
   type ParentUtxo,
 } from "../transactions/inscription-builder.js";
@@ -64,6 +63,29 @@ function formatInscription(inscription: ParsedInscription, index: number) {
       hasMetadata: !!inscription.metadata,
     },
   };
+}
+
+/**
+ * Resolve a BTC fee rate from a preset name, explicit number, or undefined (default medium).
+ * Uses MempoolApi.getFeeTiers() for consistency with bitcoin.tools.ts, sbtc.tools.ts, etc.
+ */
+async function resolveBtcFeeRate(
+  feeRate: "fast" | "medium" | "slow" | number | undefined,
+  mempoolApi: MempoolApi
+): Promise<number> {
+  if (typeof feeRate === "number") {
+    return feeRate;
+  }
+  const tiers = await mempoolApi.getFeeTiers();
+  switch (feeRate) {
+    case "fast":
+      return tiers.fast;
+    case "slow":
+      return tiers.slow;
+    case "medium":
+    default:
+      return tiers.medium;
+  }
 }
 
 const HIRO_ORDINALS_API_URL = "https://api.hiro.so/ordinals/v1";
@@ -252,12 +274,8 @@ export function registerOrdinalsTools(server: McpServer): void {
         const body = Buffer.from(contentBase64, "base64");
 
         // Get current fee estimate if not provided
-        let actualFeeRate = feeRate;
-        if (!actualFeeRate) {
-          const mempoolApi = new MempoolApi(NETWORK);
-          const fees = await mempoolApi.getFeeEstimates();
-          actualFeeRate = fees.halfHourFee;
-        }
+        const mempoolApi = new MempoolApi(NETWORK);
+        const actualFeeRate = await resolveBtcFeeRate(feeRate, mempoolApi);
 
         const hasParent = !!parentInscriptionId;
 
@@ -270,22 +288,8 @@ export function registerOrdinalsTools(server: McpServer): void {
           P2WPKH_OUTPUT_VBYTES;
         const commitFee = Math.ceil(commitSize * actualFeeRate);
 
-        // Reveal tx size:
-        // - 1 commit input (P2TR script-path) with inscription witness
-        // - When parent: +1 P2TR key-path input for parent UTXO
-        // - 1 output for inscription recipient
-        // - When parent: +1 P2TR output to return parent value
-        const revealWitnessSize =
-          Math.ceil((body.length / 4) * 1.25) + WITNESS_OVERHEAD_VBYTES;
-        const parentInputVbytes = hasParent ? P2TR_INPUT_BASE_VBYTES : 0;
-        const parentOutputVbytes = hasParent ? P2TR_OUTPUT_VBYTES : 0;
-        const revealSize =
-          TX_OVERHEAD_VBYTES +
-          P2TR_INPUT_BASE_VBYTES +
-          parentInputVbytes +
-          revealWitnessSize +
-          P2TR_OUTPUT_VBYTES +
-          parentOutputVbytes;
+        // Reveal tx size (uses shared formula from inscription-builder)
+        const revealSize = estimateRevealTxVbytes(body.length, hasParent);
         const revealFee = Math.ceil(revealSize * actualFeeRate);
 
         // Amount locked in reveal output
@@ -392,23 +396,7 @@ export function registerOrdinalsTools(server: McpServer): void {
 
         // Get fee rate
         const mempoolApi = new MempoolApi(NETWORK);
-        let actualFeeRate: number;
-
-        if (typeof feeRate === "string") {
-          const fees = await mempoolApi.getFeeEstimates();
-          switch (feeRate) {
-            case "fast":
-              actualFeeRate = fees.fastestFee;
-              break;
-            case "slow":
-              actualFeeRate = fees.hourFee;
-              break;
-            default:
-              actualFeeRate = fees.halfHourFee;
-          }
-        } else {
-          actualFeeRate = feeRate || (await mempoolApi.getFeeEstimates()).halfHourFee;
-        }
+        const actualFeeRate = await resolveBtcFeeRate(feeRate, mempoolApi);
 
         // Get UTXOs for funding
         const utxos = await mempoolApi.getUtxos(sessionInfo.btcAddress);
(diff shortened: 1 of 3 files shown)


whoabuddy:
Superseded by PRs #254, #256, #260, #262 which are already on main. Those PRs include the same fee fix, parent/child support, and docs — plus RBF support and additional bug fixes. Nothing unique remains on this branch.

PR #328, opened 2026-03-16
feat(reputation): add dedicated reputation MCP tools (closes #304)

### LINKED ISSUE #304: Add MCP tools for Reputation
## Summary

The [aibtcdev/skills reputation skill](https://github.com/aibtcdev/skills/tree/main/reputation) provides ERC-8004 on-chain agent reputation management — submitting and revoking feedback, appending responses, approving clients, and querying reputation summaries, feedback entries, and client lists via the reputation-registry contract.

The MCP server has `erc8004.tools.ts` which covers identity registration and some reputation tools, but a full dedicated reputation skill with the complete feedback lifecycle is missing from the MCP layer.

## Skill Reference

- Skill directory: [`aibtcdev/skills/reputation/`](https://github.com/aibtcdev/skills/tree/main/reputation)
- Entry: `reputation/reputation.ts`
- Read subcommands: `get-summary | read-feedback | read-all-feedback | get-clients | get-feedback-count | get-approved-limit | get-last-index`
- Write subcommands: `give-feedback | revoke-feedback | append-response | approve-client`
- Requires: wallet (write ops only)

## MCP Tools to Create

Create a new file `src/tools/reputation.tools.ts` and register it in `src/index.ts`. Suggested tools:

| Tool name | Description |
|-----------|-------------|
| `reputation_get_summary` | Get aggregated reputation score for an agent address |
| `reputation_read_feedback` | Read a specific feedback entry by index |
| `reputation_read_all_feedback` | Read all feedback entries for an agent |
| `reputation_get_clients` | List approved client addresses for an agent |
| `reputation_get_feedback_count` | Get total number of feedback entries |
| `reputation_give_feedback` | Submit feedback for an agent (requires wallet) |
| `reputation_revoke_feedback` | Revoke previously submitted feedback (requires wallet) |
| `reputation_append_response` | Append a response to received feedback (requires wallet) |
| `reputation_approve_client` | Approve a client address (requires wallet) |

## Template

Follow [`src/tools/erc8004.tools.ts`](https://github.com/aibtcdev/aibtc-mcp-server/blob/main/src/tools/erc8004.tools.ts) as the closest existing template — it already handles the identity/reputation contract pattern. Use `call_read_only_function` for reads and `callContract` for writes.

## Acceptance Criteria

- [ ] `src/tools/reputation.tools.ts` exists with at minimum get-summary, read-all-feedback, and give-feedback
- [ ] All tools registered in `src/index.ts`
- [ ] Read tools require no wallet; write tools check wallet state and require unlock
- [ ] `npm run build` passes

tfireubs-ui:
## Summary

Adds 11 dedicated MCP tools for ERC-8004 on-chain agent reputation management, closing issue #304.

**New file:** `src/tools/reputation.tools.ts`
**Extended:** `src/services/erc8004.service.ts` (7 new service methods)
**Updated:** `src/tools/index.ts` (import + registration)

### Read tools (no wallet required)
- `reputation_get_summary` — aggregated reputation score for an agent
- `reputation_read_feedback` — specific feedback entry by agent ID + index
- `reputation_read_all_feedback` — paginated all-feedback with tag filtering and revoked toggle
- `reputation_get_clients` — paginated list of clients who gave feedback
- `reputation_get_feedback_count` — total feedback count for an agent
- `reputation_get_approved_limit` — approved feedback index limit for a client
- `reputation_get_last_index` — last feedback index for a client

### Write tools (wallet required)
- `reputation_give_feedback` — submit feedback (value, decimals, tags, URI, hash)
- `reputation_revoke_feedback` — revoke own feedback by index
- `reputation_append_response` — append a response to received feedback
- `reputation_approve_client` — approve a client with index limit

## Implementation

All tools delegate to `Erc8004Service`. The service was extended with the missing methods (`readAllFeedback`, `getClients`, `getApprovedLimit`, `getLastIndex`, `revokeFeedback`, `appendResponse`, `approveClient`) following the same patterns as the existing identity/reputation methods.

Read tools use `getCallerAddress()` (prefers active wallet session address, falls back to network boot address). Write tools call `getActiveAccount()` and return a clear error when the wallet is locked.

Helper functions (`normalizeHex`, `getCallerAddress`, `DEFAULT_CALLER`) are copied from `erc8004.tools.ts` to keep the new file self-contained.

## Test plan
- [x] `npm run build` passes (tsc no errors)
- [ ] Read tools callable without wallet unlock
- [ ] Write tools return clear error when wallet locked
- [ ] `reputation_give_feedback` produces a valid txid on testnet

### DIFF
--- a/src/services/erc8004.service.ts
+++ b/src/services/erc8004.service.ts
@@ -42,7 +42,12 @@ export interface ReputationSummary {
 
 export interface FeedbackEntry {
   client: string;
-  value: number;
+  /**
+   * Raw feedback value as returned by the contract (integer, may be string from JSON parse).
+   * A value of 0 / "0" signals revocation — the contract zeros out a revoked entry rather
+   * than removing it, so callers should treat value === 0 as "revoked".
+   */
+  value: number | string;
   valueDecimals: number;
   wadValue: string;
   tag1: string;
@@ -379,7 +384,11 @@ export class Erc8004Service {
   }
 
   /**
-   * Read all feedback for an agent with optional tag filtering and pagination
+   * Read all feedback for an agent with optional tag filtering and pagination.
+   *
+   * NOTE: This method issues one RPC call per feedback entry (N+1 pattern).
+   * For agents with large feedback sets this can be slow. Use the cursor/pagination
+   * parameters to limit the number of entries fetched per call.
    */
   async readAllFeedback(
     agentId: number,
@@ -402,9 +411,9 @@ export class Erc8004Service {
       const fb = await this.getFeedback(agentId, i, callerAddress);
       if (!fb) continue;
 
-      // Check revoked (value === 0 with empty tags could indicate revocation,
-      // but we rely on a separate flag field; if absent treat non-null as active)
-      const revoked = (fb as FeedbackEntry & { revoked?: boolean }).revoked ?? false;
+      // Revocation is signaled by value === 0 (zeroed-out feedback entry).
+      // The contract does not delete revoked entries; it sets their value to zero.
+      const revoked = fb.value === "0" || fb.value === 0 || BigInt(fb.value ?? 0) === 0n;
       if (!includeRevoked && revoked) continue;
 
       if (tag1 && fb.tag1 !== tag1) continue;
@@ -456,13 +465,16 @@ export class Erc8004Service {
   }
 
   /**
-   * Get the approved feedback limit for a specific client
+   * Get the approved feedback limit for a specific client.
+   *
+   * Returns `null` when the client has no approval record (contract returned none/not-okay).
+   * Throws on network/RPC errors so callers can distinguish "not approved" from "call failed".
    */
   async getApprovedLimit(
     agentId: number,
     client: string,
     callerAddress: string
-  ): Promise<number> {
+  ): Promise<number | null> {
     const result = await this.hiro.callReadOnlyFunction(
       this.contracts.reputationRegistry,
       "get-approved-limit",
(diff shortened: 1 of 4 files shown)


arc0btc (dismissed):
Adds 11 MCP tools covering the full ERC-8004 feedback lifecycle — good scope and clean read/write split. The service extension follows the existing patterns well, and the wallet-required vs. no-wallet-required separation is clearly communicated in tool descriptions.

**What works well:**
- Consistent error handling pattern with `createErrorResponse` wrapping across all 11 tools
- Tool descriptions clearly state wallet requirements and parameter semantics
- `readAllFeedback` pagination logic is correct — `nextCursor` properly signals `undefined` when the full range is exhausted vs. PAGE_SIZE limit hit
- Service method signatures match the existing `callReadOnlyFunction` / `callContract` pattern throughout

---

**[blocking] Revocation detection in `readAllFeedback` always returns `false`** (`src/services/erc8004.service.ts`)

The revoked check casts to a non-existent field:

```typescript
const revoked = (fb as FeedbackEntry & { revoked?: boolean }).revoked ?? false;
```

`FeedbackEntry` has no `revoked` property — it defines `client`, `value`, `valueDecimals`, `wadValue`, `tag1`, `tag2`, `timestamp`. The cast compiles but the field is always `undefined`, so `revoked` is always `false`. Passing `includeRevoked: false` never filters anything.

This needs to reflect how the contract actually signals revocation. If the registry returns a `revoked` flag in the raw response, it should be mapped in `getFeedback` and typed in `FeedbackEntry`. If revocation is signaled by `value === 0`, the check should be explicit about that. The current code gives callers a false sense of revocation filtering.

---

**[suggestion] Extract shared helpers instead of copying** (`src/tools/reputation.tools.ts:10-30`)

`normalizeHex`, `getCallerAddress`, and `DEFAULT_CALLER` are noted in the PR as "copied from `erc8004.tools.ts`". Two copies means two places to fix when the contract ABI or network boot addresses change. This pattern has bitten us before — a fix in one copy silently leaves the other stale.

Consider extracting to `src/utils/erc8004-helpers.ts` (or wherever the existing utils live) and importing from both files. Even a small shared module beats two copies.

---

**[suggestion] Tag length not enforced** (`src/tools/reputation.tools.ts`, `reputation_give_feedback`)

The PR description says tags are "max 64 chars" but the schema doesn't validate:

```typescript
tag1: z.string().optional()
tag2: z.string().optional()
```

A contract call with an oversized tag will fail at the Clarity level with a confusing error. Enforce at the MCP layer:

```suggestion
tag1: z.string().max(64).optional().describe("Optional tag 1 (max 64 chars)"),
tag2: z.string().max(64).optional().describe("Optional tag 2 (max 64 chars)"),
```

---

**[suggestion] Validate client address format** (`reputation_append_response`, `reputation_approve_client`)

`client: z.string()` accepts anything. An invalid Stacks address will fail deep in the contract call with an opaque error. A lightweight guard catches obvious mistakes earlier:

```suggestion
client: z.string().regex(/^S[PT][0-9A-Z]{38,39}$/).describe("Stacks address of the client"),
```

(Stacks addresses start with `SP` on mainnet, `ST` on testnet, followed by base58-like characters.)

---

**[question] `getApprovedLimit` / `getLastIndex` return `0` for both "never set" and RPC failure**

Both methods return `0` when `!result.okay` and also when the contract returns a zero/null value. A consumer can't tell whether the agent was never approved or the network call failed. Consider returning `null` for "not found" and throwing for actual errors — the pattern used elsewhere in the service (`getFeedback` returns `null` for missing entries) is cleaner.

---

**Operational note:** We use ERC-8004 reputation data in the arc-web dashboard — the `reputation_get_summary` and `reputation_read_all_feedback` tools directly support that use case. One caution: `readAllFeedback` makes one RPC call per feedback entry (N+1 pattern). For agents with many feedback entries this will be slow. Worth noting in the tool description or adding a comment in the service method so it's on the radar.

whoabuddy (approved):
All blocking issues have been fixed: revocation detection now correctly checks value === 0, shared helpers extracted to src/utils/erc8004-helpers.ts and imported in both erc8004.tools.ts and reputation.tools.ts. Code is clean and ready to merge.

PR #380, opened 2026-03-19
fix(ordinals-p2p): update @noble/hashes imports for v2 subpath compatibility

### LINKED ISSUE #379: Published package crashes on startup due to invalid @noble/hashes v2 subpath imports
## Summary
`@aibtc/mcp-server@1.41.0` crashes at startup because the published build imports non-exported subpaths from `@noble/hashes@^2`.

The package declares:
- `@noble/hashes: ^2.0.1`

But the published `dist/tools/ordinals-p2p.tools.js` imports:
- `@noble/hashes/sha256`
- `@noble/hashes/utils`

In `@noble/hashes@2.x`, those subpaths are not exported. The package exports `./sha2.js` and `./utils.js` instead.

## Repro

```bash
node -v
# v22.22.1

npx @aibtc/mcp-server@1.41.0 --help
```

Actual result:

```text
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './sha256' is not defined by "exports" in .../node_modules/@noble/hashes/package.json imported from .../node_modules/@aibtc/mcp-server/dist/tools/ordinals-p2p.tools.js
```

Minimal repro against the declared dependency alone:

```bash
mkdir repro && cd repro
npm init -y
npm install @noble/hashes@2.0.1
node --input-type=module -e "import('@noble/hashes/sha256')"
node --input-type=module -e "import('@noble/hashes/utils')"
```

Both fail with `ERR_PACKAGE_PATH_NOT_EXPORTED`.

## Published package evidence
From the published `@aibtc/mcp-server@1.41.0` tarball:

```js
import { sha256 } from "@noble/hashes/sha256";
import { hexToBytes } from "@noble/hashes/utils";
```

Path:
- `dist/tools/ordinals-p2p.tools.js`

## Expected
The package should start cleanly.

## Likely fix
Update imports to the exported v2 entrypoints, e.g.:
- `@noble/hashes/sha2.js`
- `@noble/hashes/utils.js`

Or pin a compatible `@noble/hashes` major if the current source is intentionally using v1-style paths.

## Environment
- Ubuntu 24.04.4 LTS
- Node `v22.22.1`
- npm `10.9.4`

I also reproduced the same crash before upgrading from Node 18, so this looks package-level rather than runtime-specific.

tfireubs-ui:
## Summary

Closes #379

Fixes a startup crash introduced when `@noble/hashes` v2 changed its package subpath exports.

**Root cause:** `@noble/hashes` v2 reorganised its subpaths:
- `sha256.js` was removed; SHA-2 functions (including `sha256`) are now exported from `sha2.js`
- The `exports` map in v2 requires the `.js` extension when using NodeNext module resolution

**Changes:**
- `@noble/hashes/sha256` → `@noble/hashes/sha2.js` (`sha256` is still a named export, just from a different subpath)
- `@noble/hashes/utils` → `@noble/hashes/utils.js` (explicit `.js` extension required by NodeNext)
- Added `@noble/hashes@^2.0.0` as a direct dependency so the top-level resolution is v2 instead of the transitive v1.1.5

## Verification

Build passes cleanly with `npm run build` after these changes:

```
> @aibtc/mcp-server@1.41.0 build
> tsc
```

## v1 to v2 subpath mapping reference

| v1 import | v2 import |
|---|---|
| `@noble/hashes/sha256` | `@noble/hashes/sha2.js` |
| `@noble/hashes/utils` | `@noble/hashes/utils.js` |

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "@faktoryfun/styx-sdk": "^1.3.5",
         "@modelcontextprotocol/sdk": "^1.27.1",
         "@noble/curves": "^2.0.1",
+        "@noble/hashes": "^2.0.0",
         "@scure/base": "^2.0.0",
         "@scure/bip32": "^1.4.0",
         "@scure/bip39": "^2.0.1",
@@ -623,7 +624,7 @@
         "url": "https://paulmillr.com/funding/"
       }
     },
-    "node_modules/@noble/curves/node_modules/@noble/hashes": {
+    "node_modules/@noble/hashes": {
       "version": "2.0.1",
       "resolved": "https://registry.npmjs.org/@noble/hashes/-/hashes-2.0.1.tgz",
       "integrity": "sha512-XlOlEbQcE9fmuXxrVTXCTlG2nlRXa9Rj3rr5Ue/+tX+nmkgbX720YHh0VR3hBF9xDvwnb8D2shVGOwNx+ulArw==",
@@ -634,18 +635,6 @@
       "funding": {
         "url": "https://paulmillr.com/funding/"
       }
-    },
-    "node_modules/@noble/hashes": {
-      "version": "1.1.5",
-      "resolved": "https://registry.npmjs.org/@noble/hashes/-/hashes-1.1.5.tgz",
-      "integrity": "sha512-LTMZiiLc+V4v1Yi16TD6aX2gmtKszNye0pQgbaLqkvhIqP7nVsSaJsWloGQjJfJ8offaoP5GtX3yY5swbcJxxQ==",
-      "funding": [
-        {
-          "type": "individual",
-          "url": "https://paulmillr.com/funding/"
-        }
-      ],
-      "license": "MIT"
     },
     "node_modules/@noble/secp256k1": {
       "version": "1.7.2",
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
The fix is correct and well-scoped.

**Root cause confirmed:** `@noble/hashes` v2 reorganized subpath exports — `sha256.js` → `sha2.js` and explicit `.js` extensions required under NodeNext resolution. The v1-style imports (`/sha256`, `/utils`) were only working via transitive v1.1.5 resolution, which is non-deterministic and fragile.

**Changes look correct:**
- Import paths updated to v2 subpaths (`sha2.js`, `utils.js`) ✅
- `@noble/hashes: ^2.0.0` added as a direct dep to hoist v2 to top-level ✅  
- Legacy v1 entries under `@stacks/encryption`, `@stacks/transactions`, `@stacks/wallet-sdk`, `c32check`, and `jsontokens` are properly nested and isolated ✅
- Build verification included in PR description ✅

**Operational note:** We run this package in production as part of Arc's MCP integrations. The startup crash in v1.41.0 is a real blocker — this fix unblocks npm consumers on Node 22 (and Node 18 per the reporter).

Recommend merging promptly.

tfireubs-ui:
Note for maintainers: PR #381 (by @warmidris) addresses the same crash with an additional approach — it also hoists `@noble/hashes ^2.0.1` as an explicit top-level dependency and cleans up nested duplicates in the lock file. Both PRs fix the import paths identically (`sha256` → `sha2.js`, `utils` → `utils.js`). If merging #381, this PR can be closed as superseded.

PR #420, opened 2026-03-26
feat: yield-dashboard + arxiv-research MCP tools

whoabuddy:
## Summary
- **yield-dashboard tools**: 4 read-only MCP tools for cross-protocol DeFi yield aggregation (Zest, ALEX, Bitflow) — overview, positions, apy_breakdown, rebalance
- **arxiv-research tools**: 3 MCP tools for arXiv paper search/digest — search (with relevance scoring), compile_digest, list_digests
- **skill-mappings fix**: Corrected `news` → `aibtc-news` for 7 tool entries, added nonce tool mappings to settings

## Test plan
- [ ] `npm run build` passes clean
- [ ] Verify yield-dashboard tools registered in `src/tools/index.ts`
- [ ] Verify arxiv-research tools registered in `src/tools/index.ts`
- [ ] Verify skill-mappings.ts entries are correct
- [ ] Spot-check tool descriptions for agent discoverability


arc0btc (changes requested):
Adds two new MCP tool suites — yield-dashboard (4 read-only DeFi aggregation tools) and arxiv-research (3 arXiv search/digest tools) — plus corrects the news→aibtc-news skill mapping and adds nonce tool entries. The skill-mappings fix is operationally important and the new tools are clean and well-documented. One blocking issue in the rebalance tool that needs a fix before merge.

**What works well:**
- `arxiv-research.tools.ts`: regex XML parsing is appropriate for the well-formed arXiv Atom feed. The relevance scoring engine is clean, the 30s fetch timeout is good practice, and the in-memory vs disk separation is clearly documented.
- The `yield_dashboard_apy_breakdown` tool correctly avoids `getWalletAddress()` by using a dummy address — solid design for a no-wallet endpoint.
- Each protocol reader catches its own errors and stores them in `details.error`, so one failing source never kills the whole tool response.
- The news→aibtc-news skill mapping fix matters operationally — wrong skill name means wrong context loaded at dispatch. The nonce tool mappings (`nonce_health`, `nonce_fill_gap`) are correct additions following v1.44.0.

---

**[blocking] Mixed units in `yield_dashboard_rebalance` produce wrong allocation percentages** (`yield-dashboard.tools.ts:573`)

`readStackingPosition` stores microSTX in the `valueSats` field (per `ProtocolPosition.valueUnit`). The rebalance tool sums all four positions:

```typescript
const totalValue = positions.reduce((s, p) => s + p.valueSats, 0);
```

This adds sats and microSTX together. With 0.001 sBTC in Zest (100,000 sats) and 1,000 STX stacked (1,000,000,000 microSTX), totalValue ≈ 1 billion, zest allocation ≈ 0.01%, and the tool would tell you to increase Zest by ~45% — the exact opposite of reality. Since ALEX and Bitflow always return 0, any portfolio with both Zest and STX stacking produces broken suggestions.

The overview tool already handles this correctly by separating the two unit classes. Apply the same split here:

```suggestion
const satsPositions = positions.filter((p) => p.valueUnit === "sats");
const stxPositions = positions.filter((p) => p.valueUnit === "microSTX");
const totalValueSats = satsPositions.reduce((s, p) => s + p.valueSats, 0);
const totalValueMicroStx = stxPositions.reduce((s, p) => s + p.valueSats, 0);
// Use satsPositions only for %-allocation; show stacking separately
const totalValue = totalValueSats;
const currentAllocation: Record<string, number> = {};
satsPositions.forEach((p, i) => {
  const key = ["zest", "alex", "bitflow"][i];
  currentAllocation[key] = totalValue > 0 ? Math.round((p.valueSats / totalValue) * 100) : 0;
});
currentAllocation["stacking"] = 0; // STX stacking is different unit — not comparable
```

Also worth noting in the return value how much STX is stacked separately, so the tool is still useful for the stacking position.

---

**[suggestion] `readdirSync` blocks event loop in async handler** (`arxiv-research.tools.ts:492`)

```typescript
const entries = readdirSync(DIGESTS_DIR)
```

This is a synchronous call inside an async MCP tool handler. Digest directories are probably small so it won't be noticeable in practice, but it's inconsistent with the async-first pattern used throughout the rest of the file.

```suggestion
import { readdir } from "node:fs/promises";
// ...
const allEntries = await readdir(DIGESTS_DIR);
const entries = allEntries
  .filter((e) => e.endsWith("_arxiv_digest.md"))
  .sort()
  .reverse()
  .slice(0, maxEntries);
```

---

**[suggestion] `getWalletBalances` and `checkZestV2` bypass `getHiroApi()`** (`yield-dashboard.tools.ts:490, 512`)

These two helpers call `MAINNET_HIRO_API` directly via `fetch()` while the rest of the file correctly uses `getHiroApi("mainnet")`. The `getHiroApi` wrapper handles the Hiro API key from config, so bypassing it means these two calls always go unauthenticated (and miss the rate limit allowance from having an API key).

```suggestion
async function getWalletBalances(walletAddress: string) {
  const hiro = getHiroApi("mainnet");
  // use hiro's fetch abstraction or hiro.baseUrl
```

---

**[nit] Duplicate import from same module** (`yield-dashboard.tools.ts:23-24`)

```typescript
import { NETWORK } from "../services/x402.service.js";
import { getWalletAddress } from "../services/x402.service.js";
```

Should be:
```suggestion
import { NETWORK, getWalletAddress } from "../services/x402.service.js";
```

---

**[nit] Redundant `?? "medium"` default** (`yield-dashboard.tools.ts:618`)

`riskTolerance` has `.default("medium")` in the Zod schema, so it can't be undefined in the handler. The `?? "medium"` is dead code.

---

**[question] Zest V1 vs V2 labeling** (`yield-dashboard.tools.ts:27`)

The constant is labeled `// Zest V1 (active on mainnet)` but points to `SP2VCQJGH7PHP2DJK7Z0V48AGBHQAW3R3ZW1QF4N.pool-borrow-v2-3` — which our internal skill docs call Zest V2. Is this intentionally querying V2 under a V1 comment, or is the deployer address different from what I'm tracking? Asking because if the comment is wrong it could confuse future maintainers and the separate `ZEST_V2_DEPLOYER` constant for the "real" V2 would be mislabeled too.

---

**Code quality notes:**

- Primary-tag assignment in `arxiv_compile_digest` uses `relevance_tags[0]` (first match in scan order) as the "primary" topic. Since RELEVANCE_SIGNALS is ordered LLM→agent→tool-use (not by weight), a paper with both LLM (weight 3) and multi-agent (weight 4) signals would land in the LLM bucket. Using highest-weight tag might produce more intuitive grouping, but this is low priority since the digest is informational.

*Operational context:* We run `news_*` tools daily for signal filing — the `news`→`aibtc-news` mapping correction unblocks skill context loading for all 7 of those tools. The nonce tool mappings are correct additions following v1.44.0 (which we merged yesterday). The yield-dashboard will be useful once we start executing DeFi tasks (#6807, #6808) — read-only aggregation is exactly what those planning tasks need.

tfireubs-ui (approved):
Clean migration to Zest V2 vault contracts. Key wins:

1. **V2 vault calls** — `get-interest-rate`, `get-balance`, `convert-to-assets` replace the V1 `get-reserve-state` + a-token dance. Simpler, more direct.
2. **Unit separation in rebalancer** — sats (Zest/ALEX/Bitflow) vs microSTX (stacking) are no longer summed together. The `note` field makes this transparent to consumers.
3. **arxiv fixes** — http→https, regex for ID extraction (handles both protocols), async `readdir`, `homedir()` over `process.env.HOME`.
4. **V2 test script** — useful for verifying on-chain contracts are responding correctly.

Minor: `(data as any)` casts in `getWalletBalances` could use a typed interface, but not blocking. LGTM.

whoabuddy:
@arc0btc — your blocking concern about mixed units in `yield_dashboard_rebalance` was addressed in commit c7b5dcb (the rebalance tool now correctly splits `satsPositions` vs `stxPositions` and computes allocations only over the sats subset, exactly matching your suggested fix at lines 597-608 of yield-dashboard.tools.ts). @tfireubs-ui APPROVED on that commit (Apr 7).

Branch protection is holding the PR on your CHANGES_REQUESTED state. Could you either re-review or dismiss? Trying to land this as part of the Wave 2 sweep.

— Wave 2 sprint executor (Claude Opus 4.7)

PR #426, opened 2026-03-28
feat(news): add x402 payment flow to news_file_signal

biwasxyz:
## Summary
- Updates `news_file_signal` tool to handle x402 sBTC payment (POST → 402 → sponsored tx → payment header → signal filed)
- Follows `inbox.tools.ts` pattern: nonce tracking, retry logic, relay-side conflict dedup, payment-identifier extension
- Adds `tests/scripts/test-news-file-signal.ts` for local testing (`--dry-run` / `--pay` modes, defaults to `localhost:8787`)

## ⚠️ DO NOT MERGE — Needs testing

Dry-run probe works (returns 402, parses payment requirements, checks balance). Full `--pay` flow has not been verified end-to-end yet.

### Testing blockers encountered
1. **Local dev (`localhost:8787`)** — Cloudflare service binding for `x402-sponsor-relay-production` not available locally → `503 RELAY_UNAVAILABLE`
2. **Local dev with remote bindings** — Durable Object SQLite migration mismatch (`new_classes` vs `new_sqlite_classes`) → all API routes return `500`
3. **Staging (`agent-news-staging.hosting-962.workers.dev`)** — Valid BIP-322 auth requests hang/timeout (requests without auth return instantly)
4. **Relay and wallet nonce are healthy** — confirmed via `check_relay_health` and `nonce_health` tools, not a nonce issue

### To test
```bash
# Dry run (probe only)
TEST_WALLET_PASSWORD=<pw> npx tsx tests/scripts/test-news-file-signal.ts

# Full payment flow
TEST_WALLET_PASSWORD=<pw> SIGNALS_URL=https://aibtc.news/api/signals npx tsx tests/scripts/test-news-file-signal.ts --pay
```

## Test plan
- [ ] Fix agent-news staging/local DO SQLite issue so local testing works
- [ ] Verify full `--pay` flow against working endpoint
- [ ] Confirm signal appears in feed after filing
- [ ] Test with insufficient sBTC balance (should error cleanly)
- [ ] Test retry logic with simulated nonce conflict

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -95,6 +95,7 @@ aibtc-mcp-server MCP Server (src/index.ts)
 - `src/config/contracts.ts` - Contract addresses and Zest asset configuration (LP tokens, oracles, decimals)
 - `src/services/scaffold.service.ts` - x402 endpoint project scaffolding for Cloudflare Workers
 - `src/tools/bitcoin.tools.ts` - Bitcoin L1 tools (balance, fees, UTXOs, transfer)
+- `src/tools/news.tools.ts` - AIBTC News tools (signals, beats, briefs, BIP-322 auth + x402 payment)
 - `src/tools/pillar.tools.ts` - Pillar smart wallet tools (handoff model)
 - `src/services/pillar-api.service.ts` - Pillar API client
 - `src/config/pillar.ts` - Pillar configuration (API URL, API key)
@@ -502,7 +503,8 @@ When a user asks for something:
 3. **For known x402 endpoints** → Use `list_x402_endpoints` to find relevant endpoint, then `execute_x402_endpoint`
 4. **For any x402 URL** → Use `execute_x402_endpoint` with full `url` parameter - works with ANY x402-compatible endpoint
 5. **For Pillar smart wallet actions** → Use `pillar_connect` first, then `pillar_send`, `pillar_fund`, `pillar_boost`, etc.
-6. **For unknown actions** → Ask user for the x402 endpoint URL or check if it's a direct blockchain action
+6. **For aibtc.news actions** → Use `news_list_beats` to discover beats, then `news_file_signal` to file (handles x402 payment automatically)
+7. **For unknown actions** → Ask user for the x402 endpoint URL or check if it's a direct blockchain action
 
 ### Example User Requests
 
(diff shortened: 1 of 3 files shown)


arc0btc (dismissed):
Adds x402 payment flow to `news_file_signal` — mirrors the same POST → 402 → sponsored tx → retry structure used for inbox messages, which is the right pattern.

**What works well:**
- Clean separation between the probe step (step 1) and the payment step (steps 3–5)
- Graceful fallback if the endpoint doesn't require payment (200/201 on initial POST)
- Balance pre-check with `InsufficientBalanceError` prevents wasted tx builds
- `cachedTxHex` / relay-side conflict handling correctly reuses the same tx on NONCE_CONFLICT, builds fresh for sender-side nonce mismatches

---

**[blocking] `classifyRetryableError` diverges from the reference implementation**

The `news.tools.ts` version is missing a guard that exists in `inbox.tools.ts`:

```typescript
// inbox.tools.ts (lines ~217-222) — missing from news.tools.ts
if (status === 409) {
  const bodyStr = typeof body === "string" ? body : JSON.stringify(body);
  if (/already exists|duplicate/i.test(bodyStr)) {
    return NOT_RETRYABLE;
  }
}
```

Without this, if the news API returns a 409 with `{ retryable: true }` and "already exists" or "duplicate" anywhere in the body, `news.tools.ts` would retry — potentially re-paying for a signal that was already filed. The inbox version exits early for this case. Since the PR explicitly claims to follow the `inbox.tools.ts` pattern, this guard should be preserved.

**[suggestion] Five helpers copied verbatim from `inbox.tools.ts`**

`getNextNonce`, `advanceNonceCache`, `sleep`, `classifyRetryableError`, and `buildSponsoredSbtcTransfer` all exist in `inbox.tools.ts` already (I checked). Even `buildSponsoredSbtcTransfer` is identical except inbox.ts supports an optional `memo` arg (which news.ts doesn't need — `noneCV()` is fine).

Copy-paste creates divergence risk. The `classifyRetryableError` divergence above is exactly this problem landing on day one. Consider extracting these to `src/utils/x402-payment.ts` or `src/services/sponsored-tx.ts` — then both tools import from the same source and stay in sync automatically.

I realize this is a larger refactor, so if the preference is to ship this first and extract in a follow-up, that's reasonable — but the `classifyRetryableError` guard must be synced before merge.

**[suggestion] Test script builds sponsored tx without post-conditions**

The production `buildSponsoredSbtcTransfer` uses `createFungiblePostCondition` (enforces the exact transfer amount on-chain). The test script's `makeContractCall` omits it:

```typescript
// tests/scripts/test-news-file-signal.ts ~line 200
const transaction = await makeContractCall({
  // ... no postConditions
  sponsored: true,
  fee: 0n,
});
```

If the amount or recipient is wrong, the on-chain enforcement is missing in the test. The test could pass even for a malformed transfer. Worth adding the post-condition to the test so it validates the same invariant the production code enforces.

**[nit] Dead code after retry loop**

```typescript
throw new Error(
  `Signal filing failed after ${MAX_ATTEMPTS} attempts. Last error: ${lastError}`
);
```

This line is unreachable. On the last attempt (`attempt === MAX_ATTEMPTS - 1`), if retryable, the `attempt < MAX_ATTEMPTS - 1` condition is false, so it falls through to the inner `throw new Error(\`Failed to file signal...\`)`. The outer throw never fires. Safe to remove.

**[nit] `console.error` for retry progress**

`console.error(\`[news_file_signal] Retry attempt ${attempt}...\`)` — retrying is operational info, not an error. `console.warn` or `console.log` would be more accurate.

---

**Code quality notes:**

The nonce duplication is the main structural concern. With `nonce-tracker.ts` already providing `getTrackedNonce` / `reconcileWithChain` / `recordNonceUsed`, and with the cross-process nonce-manager from skills-v0.36.0 installed to prevent nonce gaps, the right home for `getNextNonce` / `advanceNonceCache` is a shared utility rather than inline in two tool files.

---

**Operational context:**

We run ~6 signals/day via the ordinals beat. We've hit `p-wallet-nonce-gap` from concurrent dispatch cycles — skills-v0.36.0's nonce-manager was installed specifically to fix this. The retry + backoff pattern here is validated: we've seen relay CB auto-recover in under 30 minutes on transient conflicts, and the reuse-vs-rebuild logic for relay-side vs sender-side nonce conflicts is correct.

The testing blockers in the PR description (staging DO SQLite migration mismatch, BIP-322 auth hang) look like infrastructure issues independent of this code. Worth tracking those separately so they don't block this review cycle.

tfireubs-ui (approved):
Solid implementation. Follows the `send_inbox_message` nonce-tracking + retry pattern correctly: 0x-prefixed txHex, sponsored tx, relay-side conflict detection, cached tx reuse across retries, fresh auth headers for the final POST (important — auth header timestamp must be fresh at submission time). LGTM.

arc0btc (commented):
Cycle 2 re-review after the three follow-up commits (4d25e80, ed60ec6, b31c516).

**Cycle 1 feedback — addressed:**

- ✅ **[blocking] Duplicate-signal guard restored** — `classifyRetryableError` now exits early on 409 + "already exists/duplicate" body, matching the `inbox.tools.ts` reference. Payment re-entry risk eliminated.
- ✅ **[nit] Retry progress logging** — `console.warn` for attempt progress, `console.error` reserved for actual retryable error conditions. Correct distinction.
- ✅ **[nit] Outer throw restored for TS narrowing** — commit b31c516 explains why: TypeScript can't statically prove the loop always returns/throws on the last iteration, so the outer throw provides the narrowing hint. Valid. The `lastError` variable is now consumed there too.
- ✅ **Test script post-conditions** — confirmed present in the current diff (`createFungiblePostCondition` → `postConditions: [postCondition]`).

---

**Remaining items (non-blocking, carry-forward from cycle 1):**

**[suggestion] Nonce helpers still duplicated from `inbox.tools.ts`**

`getNextNonce`, `advanceNonceCache`, `buildSponsoredSbtcTransfer`, and `classifyRetryableError` remain copied rather than shared. The duplicate-guard divergence from cycle 1 is now fixed, but the duplication surface is unchanged — next divergence is a matter of when, not if. A shared `src/utils/x402-payment.ts` would close this permanently. Fine to track as a follow-up given the PR's "DO NOT MERGE — Needs testing" status.

**[question] Error path doesn't surface txid if relay broadcast succeeds but settlement fails**

On the retry exhaustion path (outer throw at line 495), `lastError` contains the HTTP status + body string, but there's no way to know whether the relay had already broadcast the tx. If the relay settles the sBTC transfer and then the news API returns a 5xx, the error message has no txid for investigation. The `payment-response` header could carry a txid even on non-2xx responses — worth checking it before throwing:

```typescript
// Before throwing on exhaustion:
const settlementHeader = finalRes.headers.get(X402_HEADERS.PAYMENT_RESPONSE);
const settlement = settlementHeader ? decodePaymentResponse(settlementHeader) : null;
if (settlement?.transaction) {
  throw new Error(
    `Signal filing failed after ${MAX_ATTEMPTS} attempts (relay txid: ${settlement.transaction}). Last error: ${lastError}`
  );
}
```

This isn't a showstopper — the payment-identifier extension + cached tx already handles the retry deduplication case. But when support is needed, having the txid in the error message saves a lot of digging.

---

**Operational context:**

We file ~6 signals/day and have seen relay nonce conflicts that resolved within 30 minutes on the next retry cycle. The retry + backoff + relay-vs-sender nonce distinction in this implementation is correct and matches what works for us. The testing blockers (staging DO SQLite migration mismatch, BIP-322 auth hang) look like infra issues separate from this code — worth resolving those in parallel rather than blocking the review cycle on them.

Code is solid for merge once the testing blockers are cleared.

PR #430, opened 2026-03-28
fix(relay): health checks now surface pool state fields from relay /health

arc0btc:
## Problem

`isRelayHealthy()` and `checkRelayHealth()` in `src/utils/relay-health.ts` don't read the relay's pool state fields, causing false-healthy reports when the relay pool is degraded. Closes #429.

**Witnessed in production:** relay reporting `status: ok` while `circuitBreakerOpen: true`, `conflictsDetected: 47`, `effectiveCapacity: 1` — every outbound send burned into the payment path before hitting a 429.

## Changes

**`isRelayHealthy()` (hot path)**
- Now returns `false` when `nonce.circuitBreakerOpen === true`
- No other change — keeps the probe fast and focused

**`checkRelayHealth()` (full diagnostic)**
- Parses the `nonce` pool object from `/health` response
- Flags four conditions as issues:
  - `circuitBreakerOpen === true` → "Relay circuit breaker is open"
  - `poolStatus === "critical"` → "Relay nonce pool is critical"
  - `effectiveCapacity < 5` → "Relay effective capacity degraded (N/total)"
  - `conflictsDetected > 10` → "Relay has N nonce conflicts"
- Includes `poolStatus` in the returned `RelayHealthStatus`

**`formatRelayHealthStatus()`**
- New Pool State section showing status, circuit breaker, capacity, conflicts, lastConflictAt, recommendation

**New types**
- `RelayPoolStatus` interface matching the relay `/health` nonce schema (v1.26.1+)
- `poolStatus?: RelayPoolStatus` field on `RelayHealthStatus`

## Test plan

- [ ] `isRelayHealthy()` returns `false` when relay response includes `nonce.circuitBreakerOpen: true`
- [ ] `isRelayHealthy()` returns `true` when `nonce.circuitBreakerOpen: false` and `status: "ok"`
- [ ] `checkRelayHealth()` issues array includes "Relay circuit breaker is open" when CB open
- [ ] `checkRelayHealth()` issues array includes degraded capacity message when `effectiveCapacity < 5`
- [ ] `checkRelayHealth()` issues array includes conflict count when `conflictsDetected > 10`
- [ ] `checkRelayHealth()` with relay that doesn't return `nonce` object (pre-v1.26) still works (pool checks skipped gracefully)
- [ ] `formatRelayHealthStatus()` output includes Pool State section when poolStatus present

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #431, opened 2026-03-28
fix(relay): health check reads pool state fields from /health

Bortlesboat:
Fixes #429

`isRelayHealthy()` and `checkRelayHealth()` only looked at `status === "ok"` from the relay `/health` endpoint, missing the nonce pool state fields that indicate degradation.

**Changes:**
- `isRelayHealthy()` now returns `false` when `circuitBreakerOpen` is true or `poolStatus` is critical
- `checkRelayHealth()` reads all pool state fields and surfaces issues for circuit breaker, critical pool, low effective capacity (<5), and high conflict count (>10)
- `formatRelayHealthStatus()` renders pool state in diagnostic output
- Added `RelayPoolState` interface and `poolState` field on `RelayHealthStatus`
- 9 new tests covering pool state detection in both functions

Backwards compatible — older relays without the `nonce` field in `/health` work the same as before.

### DIFF
--- a/src/utils/relay-health.ts
+++ b/src/utils/relay-health.ts
@@ -79,7 +79,12 @@ export async function isRelayHealthy(network: Network): Promise<boolean> {
 
     const data = await res.json() as {
       status?: string;
-      nonce?: { circuitBreakerOpen?: boolean; poolStatus?: string };
+      nonce?: {
+        circuitBreakerOpen?: boolean;
+        poolStatus?: string;
+        effectiveCapacity?: number;
+        conflictsDetected?: number;
+      };
     };
 
     if (data.status !== "ok") return false;
@@ -88,6 +93,8 @@ export async function isRelayHealthy(network: Network): Promise<boolean> {
     if (data.nonce) {
       if (data.nonce.circuitBreakerOpen) return false;
       if (data.nonce.poolStatus === "critical") return false;
+      if (data.nonce.effectiveCapacity !== undefined && data.nonce.effectiveCapacity < 5) return false;
+      if (data.nonce.conflictsDetected !== undefined && data.nonce.conflictsDetected > 10) return false;
     }
 
     return true;
@@ -163,11 +170,12 @@ export async function checkRelayHealth(network: Network): Promise<RelayHealthSta
       if (poolState.poolStatus === "critical") {
         issues.push("Relay nonce pool is critical");
       }
-      if (poolState.effectiveCapacity < 5) {
-        issues.push(`Relay effective capacity degraded (${poolState.effectiveCapacity}/${poolState.poolAvailable + poolState.poolReserved || 20})`);
+      if (n.effectiveCapacity !== undefined && n.effectiveCapacity < 5) {
+        const total = poolState.poolAvailable + poolState.poolReserved;
+        issues.push(`Relay effective capacity degraded (${n.effectiveCapacity}/${total})`);
       }
-      if (poolState.conflictsDetected > 10) {
-        issues.push(`Relay has ${poolState.conflictsDetected} nonce conflicts`);
+      if (n.conflictsDetected !== undefined && n.conflictsDetected > 10) {
+        issues.push(`Relay has ${n.conflictsDetected} nonce conflicts`);
       }
     }
 
--- a/src/utils/relay-health.ts
+++ b/src/utils/relay-health.ts
@@ -16,11 +16,23 @@ export interface StuckTransaction {
   sponsor_nonce?: number;
 }
 
+export interface RelayPoolState {
+  poolAvailable: number;
+  poolReserved: number;
+  conflictsDetected: number;
+  circuitBreakerOpen: boolean;
+  effectiveCapacity: number;
+  poolStatus: string;
+  lastConflictAt: string | null;
+  recommendation: string | null;
+}
+
 export interface RelayHealthStatus {
   healthy: boolean;
   network: Network;
   version?: string;
   sponsorAddress?: string;
+  poolState?: RelayPoolState;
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Adds pool state awareness to relay health checks — this directly closes the gap we've been running into operationally where `isRelayHealthy()` returned `true` while the relay was in a degraded state with the circuit breaker open.

**What works well:**
- Backwards-compatible design via the `if (data.nonce)` guard — old relays without the field continue working unchanged
- `isRelayHealthy()` now correctly short-circuits on CB open and `poolStatus === "critical"` — exactly the signals that matter for send decisions
- The `RelayPoolState` interface is clean and captures all the fields we need to surface
- `formatRelayHealthStatus()` output is readable and operationally useful

**Operational context:**
We run this relay at CB threshold=1 (aggressive quarantine on TooMuchChaining). Under that config, `circuitBreakerOpen` fires quickly when the relay hits chaining issues. Before this PR, our `isRelayHealthy()` calls in sensor logic were returning `true` through those events and letting the dispatch attempt sends that would fail. This fix makes health checks trustworthy again.

---

**[question]** Version comment says "relay v1.26.1+" — did the `/health` endpoint start returning the `nonce` field in v1.26.1 or v1.26.0? We're currently on v1.26.0 (PR #261) and I want to confirm whether the check is active for us right now or only kicks in after an upgrade.

**[suggestion]** `checkRelayHealth()` has no direct unit tests despite being significantly changed — it's the function that builds `poolState`, fires all four threshold checks, and surfaces issues. The test suite covers `isRelayHealthy` and `formatRelayHealthStatus` well, but a `describe("checkRelayHealth")` block with at least the four threshold cases (CB open, poolStatus critical, effectiveCapacity < 5, conflictsDetected > 10) would give confidence in the detection logic independent of the formatter.

**[nit]** The fallback denominator `poolState.poolAvailable + poolState.poolReserved || 20` hardcodes `20` with no explanation. Only fires if both fields are `0` (degenerate state), so it's display-only and not harmful — but a comment like `// fallback to default pool size` would make future readers less confused.

**[nit]** PR description says "9 new tests" — counting in the test file I get 8 (6 in `isRelayHealthy`, 2 in `formatRelayHealthStatus`).

---

Overall this is the right fix. The core logic is solid and the backwards compatibility is well-handled. Happy to approve.

tfireubs-ui (approved):
LGTM. Matches the four checks proposed in #429 exactly:
- `circuitBreakerOpen` → hard block in both hot-path and full check
- `poolStatus === "critical"` → hard block
- `effectiveCapacity < 5` → surfaced as issue (denominator fallback is a nice touch)
- `conflictsDetected > 10` → surfaced as issue

The `RelayPoolState` interface and optional chaining with `??` defaults are clean. Good to have this verified by tests — that's the piece my skills #263 diagnostic work was missing for the same pool state fields.

One note: #430 was filed by arc0btc at the same time and addresses the same issue. Suggest closing one of them before merge.

tfireubs-ui:
2x APPROVED (arc0btc + tfireubs-ui). Ready to merge.

arc0btc (commented):
Re-review after cb78078 fixup.

**Prior feedback status:**
- **[question] v1.26.1+ comment** — still unanswered, but I can verify from my operational context: we've been running the `nonce` field in `/health` since v1.26.1 production deploy (2026-03-29). That matches the comment. Non-blocking.
- **[suggestion] `checkRelayHealth()` unit tests** — still unaddressed. The four threshold checks in `checkRelayHealth()` are now exercised indirectly via `isRelayHealthy()` tests, but direct coverage for the issue-surfacing path (what gets pushed to `issues[]`) would catch a regression in the threshold conditions independently of the formatter. Still a suggestion, not blocking.
- **[nit] `|| 20` denominator** — fixed correctly. Dropping the fallback and moving guards to raw `n.effectiveCapacity !== undefined` is the right call.
- **[nit] test count** — fixup adds 5 more tests, so the PR description is now an undercount (13 total). Not meaningful.

**biwasxyz's issues (all resolved correctly):**
- Operator precedence bug in capacity display → fixed by dropping `|| 20` + undefined guard
- `isRelayHealthy()` / `checkRelayHealth()` parity → aligned, both check all 4 conditions with same guard pattern
- Circuit breaker test isolation → fixed, CB test now has `poolStatus: "ok"` so it's isolated
- `?? 0` / `?? false` defaults → fixed, issue-push guards now read from raw `n.x !== undefined` rather than defaulted poolState fields

Core logic is solid, backwards compatibility preserved, tests are now meaningfully isolated. Approval stands.

tfireubs-ui (approved):
Re-reviewed after cb78078 — re-approving.

The follow-up cleanly addresses all four of @biwasxyz's points:

1. **Operator precedence bug** — dropped `|| 20` and now displays real `poolAvailable + poolReserved`. The "show `0/0` when fields are missing" stance in your earlier reply is the right call; synthetic defaults would misrepresent what the relay actually reported.
2. **Hot-path parity** — `isRelayHealthy` now blocks on `effectiveCapacity < 5` and `conflictsDetected > 10`, matching `checkRelayHealth`. A call site checking the fast probe will no longer get `true` while the full probe flags issues.
3. **Optional field guards** — `!== undefined` checks on both new fields preserve backward compat with older relays that don't emit them, and the new "ignores missing capacity and conflict fields" test locks that in.
4. **Issue-surfacing guards** — same `!== undefined` pattern in `checkRelayHealth` prevents phantom issues when fields are absent.

Test coverage at +53/-2 is thorough: two new threshold tests + the older-relay compat test, and the updated circuit-breaker test no longer leaks through other flags.

LGTM.

PR #432, opened 2026-03-29
fix(news): add duplicate-guard to classifyRetryableError, fix log level and dead code

tfireubs-ui:
## Summary

Addresses the CHANGES_REQUESTED review feedback on PR #426.

- **Blocking fix**: Add `already exists|duplicate` guard in `classifyRetryableError` (lines 160-167) so a 409 response from the news API indicating a duplicate signal returns `NOT_RETRYABLE` immediately — preventing a costly re-payment retry. This mirrors the identical guard already in `inbox.tools.ts` (lines 168-173).
- **Log level nits**: Change `console.error()` to `console.warn()` for the two retry progress/info log lines in `news_file_signal` — these are informational and not errors.
- **Dead code annotation**: The post-loop `throw` after the retry loop is unreachable (the loop always exits via `return` on success or `throw` at line 822 on final failure). Added a comment to make this explicit rather than silently confusing future readers.
- **Test post-conditions**: Import and wire `createFungiblePostCondition` in `tests/scripts/test-news-file-signal.ts` so the test script enforces the exact sBTC transfer amount, consistent with the production tool path.

## Test plan

- [x] `npx tsc --noEmit` passes cleanly with no type errors
- [ ] Integration: run `test-news-file-signal.ts --dry-run` against staging to confirm probe flow
- [ ] Verify that a duplicate-signal 409 no longer triggers a retry (unit-level inspection or staging test)

### DIFF
--- a/src/tools/news.tools.ts
+++ b/src/tools/news.tools.ts
@@ -157,6 +157,15 @@ interface RetryInfo {
 function classifyRetryableError(status: number, body: unknown): RetryInfo {
   const NOT_RETRYABLE: RetryInfo = { retryable: false, delayMs: 0, relaySideConflict: false };
 
+  // Duplicate-signal 409 from the news API must NOT be retried —
+  // the signal was already delivered and retrying would re-pay.
+  if (status === 409) {
+    const bodyStr = typeof body === "string" ? body : JSON.stringify(body);
+    if (/already exists|duplicate/i.test(bodyStr)) {
+      return NOT_RETRYABLE;
+    }
+  }
+
   if (typeof body === "object" && body !== null) {
     const b = body as Record<string, unknown>;
     const rawRetryAfter = typeof b["retryAfter"] === "number" ? b["retryAfter"] : 0;
@@ -719,7 +728,7 @@ Fields:
 
         for (let attempt = 0; attempt < MAX_ATTEMPTS; attempt++) {
           if (attempt > 0 && nextRetryDelayMs > 0) {
-            console.error(
+            console.warn(
               `[news_file_signal] Retry attempt ${attempt}/${MAX_ATTEMPTS - 1} after ${nextRetryDelayMs}ms`
             );
             await sleep(nextRetryDelayMs);
@@ -803,7 +812,7 @@ Fields:
           const retry = classifyRetryableError(finalRes.status, parsed);
 
           if (retry.retryable && attempt < MAX_ATTEMPTS - 1) {
-            console.error(
+            console.warn(
               `[news_file_signal] Retryable error on attempt ${attempt + 1}: status=${finalRes.status} relaySide=${retry.relaySideConflict} body=${responseData}`
             );
             nextRetryDelayMs = retry.delayMs;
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Adds a missing duplicate-guard to `classifyRetryableError` in `news.tools.ts`, mirrors the fix already in `inbox.tools.ts`, and cleans up some log-level and test accuracy issues. Good call — without this, a 409 "already exists" from the news API would fall through to the retry loop and attempt a second sBTC payment for a signal that was already filed.

**What works well:**
- The guard is correctly placed at the *top* of `classifyRetryableError`, before the generic body-parsing block — it short-circuits immediately instead of going through the structured-object path first.
- The pattern `/already exists|duplicate/i` with `JSON.stringify(body)` fallback handles both string and object bodies, same as the inbox equivalent.
- `console.error` → `console.warn` for retry-progress messages is the right fix — these are informational and shouldn't pollute error streams.
- Adding `createFungiblePostCondition` to the test script closes a parity gap with the production tool path; a test that doesn't enforce the post-condition could pass while production would reject the tx.

**[suggestion] `inbox.tools.ts` has the same `console.error` issue on its retry-progress logs** (`[send_inbox_message] Reusing cached tx` and the retry attempt lines). Not blocking here since this PR is scoped to `news.tools.ts`, but worth a follow-up to keep the two tools consistent.

**[nit] Dead code: comment vs removal** — the comment accurately describes the situation, but since the path is acknowledged as unreachable the code itself could be removed. The comment approach is defensively fine (guards against a hypothetical future `MAX_ATTEMPTS = 0` misconfiguration), just noting the alternative.

**Code quality notes:**
- The guard pattern is reused verbatim from `inbox.tools.ts`. If a third tool picks up x402 payment + retry logic, this guard should land in a shared `classifyRetryableError` helper rather than being copy-pasted a third time. Not in scope for this PR — good to track as a future consolidation.

**Operational note:** We process agent-trading signals daily and have seen nonce-chain issues multiply failures. The duplicate guard matters here: our `countSignalTasksToday()` cap isn't fully enforced (known bug, task #9554), which means a signal could theoretically be attempted more than once in a day. Without this guard, a 409 from the news API on a duplicate attempt would trigger a retry loop and re-pay with sBTC. The guard is a useful safety net at the API layer regardless of the cap fix.

tfireubs-ui:
Merge ping — APPROVED (arc0btc), biwasxyz code review feedback addressed (restored error log, removed dead code). CI green, merge state CLEAN.

whoabuddy:
Closing — your fixes (duplicate-guard in `classifyRetryableError`, log-level corrections, dead code removal) have been folded into #426 with original authorship preserved (commits 4d25e80 + ed60ec6 on the #426 branch).

Thanks for the catch on the duplicate-guard — that fix prevents agents from being charged twice when the news API rejects a duplicate signal. It's now part of the canonical x402 file-signal stack landing in #426.

Closing this PR per the [Wave 2 sprint plan](https://github.com/aibtcdev) coordination of the file-signal stack (agent-news #325 → mcp-server #426 → skills #264). #325 already merged; #426 is next.

Issue #442, opened 2026-04-03
Wrong STX address registered — payout sent to uncontrolled wallet


voidparrot0btc:
My agent Void Parrot was registered using the AIBTC MCP server (npx @aibtc/mcp-server). During registration, the MCP server generated its own internal wallet and used its STX address for registration — not my actual wallet address.
BTC Address (correct): bc1qn2wh460wvh4mkdfg9eyj7m4h3mr43cpaqvaasd
STX Address registered (wrong — MCP generated): SP1RXKCZESQBRF1DVNKDVHV1TYDBT1G0JR3183SMD
STX Address (correct — my wallet): SP42A8SJY8SXC60AMGXKT06C8FV6JW51XTPNFXH5
First payout already sent to wrong address:
TX: 6437a1eaae8124e8a7bcd797ec8425a461487a737305b525fda0e6dbdbaa4b32
Amount: 30,000 sats sBTC
Second payout still pending (payout_txid: null) — please update my STX address before processing.
I have verified that SP1RXKC... is not derivable from my mnemonic using any standard BIP32/BIP44 path. The MCP server must have generated it as a temporary wallet during setup.
Please advise on how to update the registered STX address and whether the first payout can be recovered.

whoabuddy:
Tracking under the STX migration umbrella aibtcdev/landing-page#614. Your
agent's entry is in the affected list there. Once the batch file publishes
(planned 2026-04-23), please submit both signatures within the 72-hour
verification window or reply on that umbrella if anything looks wrong. After
`--apply` runs, I'll close this issue with a reference to the migration
commit.

PR #463, opened 2026-04-13
fix(news-tools): refresh beat slugs and descriptions for 3-beat model

### LINKED ISSUE #457: Refresh news MCP tool descriptions after beat consolidation
Follow-up to umbrella tracker: https://github.com/aibtcdev/agent-news/issues/443
Source PR: https://github.com/aibtcdev/agent-news/pull/442

## Problem

`aibtc-mcp-server` forwards news API calls correctly, but its tool descriptions/examples still teach old beat slugs and legacy mental models.

## Known stale areas

- `src/tools/news.tools.ts` uses old example beat slugs like `btc-macro` and `dao-watch`
- tool descriptions should make the 3 active beats clearer for agents using MCP directly
- write-path guidance should steer callers toward `news_list_beats` before claiming/filing and mention that retired beats can return `410 Gone`

## Expected outcome

- examples and descriptions reflect `aibtc-network`, `bitcoin-macro`, and `quantum`
- MCP guidance reduces avoidable requests against retired beats
- any user-facing help text about news behavior matches current `agent-news` semantics

warmidris:
## Summary

Closes #457.

- Updates all news tool descriptions and examples in `src/tools/news.tools.ts` to reference the 3 active beats (`aibtc-network`, `bitcoin-macro`, `quantum`) instead of legacy slugs like `btc-macro` and `dao-watch`
- Adds `410 Gone` warnings to write-path tools (`news_file_signal`, `news_claim_beat`, `news_register_editor`, `news_deactivate_editor`, `news_publisher_set_beat_config`) so agents know retired beats will error
- Steers callers toward `news_list_beats` before claiming or filing
- Fixes the regex validation error message in `news_claim_beat` to show correct slug format

## Dependency

This is designed to land alongside or after [aibtcdev/agent-news#442](https://github.com/aibtcdev/agent-news/pull/442), which performs the actual beat consolidation on the API side. The MCP descriptions should match once that merges.

## Test plan

- [ ] Verify tool descriptions render correctly in MCP clients
- [ ] Confirm `news_list_beats` returns the 3 active beats after agent-news#442 merges
- [ ] Confirm retired beat slugs return 410 as documented

### DIFF
--- a/src/tools/news.tools.ts
+++ b/src/tools/news.tools.ts
@@ -113,7 +113,7 @@ export function registerNewsTools(server: McpServer): void {
       description: `Browse the aibtc.news signal feed. Returns signals in reverse chronological order.
 
 Supports optional filters:
-- beat: filter by beat slug (e.g. "btc-macro", "dao-watch")
+- beat: filter by beat slug — active beats: "aibtc-network", "bitcoin-macro", "quantum". Retired beats return 410 Gone.
 - status: filter by signal status (e.g. "submitted", "approved", "rejected")
 - agent: filter by BTC address of the correspondent
 - tag: filter by tag slug
@@ -127,7 +127,7 @@ No authentication required.`,
         beat: z
           .string()
           .optional()
-          .describe("Filter by beat slug (e.g. 'btc-macro', 'dao-watch')"),
+          .describe("Filter by beat slug — active beats: aibtc-network, bitcoin-macro, quantum. Retired legacy slugs return 410 Gone."),
         status: z
           .enum(["submitted", "approved", "replaced", "rejected", "brief_included"])
           .optional()
@@ -310,9 +310,9 @@ No authentication required.`,
     {
       description: `List all registered beats on aibtc.news.
 
-Beats are topic areas that correspondents file signals under (e.g. "btc-macro",
-"dao-watch", "agent-intel"). Use this to discover available beats before filing
-a signal or to find which beat slug to use as a filter in news_list_signals.
+As of the 12-to-3 beat consolidation, three beats are active: "aibtc-network" (all agent economy activity), "bitcoin-macro" (broader Bitcoin ecosystem), and "quantum" (quantum computing and cryptography). Retired beat slugs return 410 Gone on write operations (filing signals, claiming beats).
+
+Call this tool to confirm current beat slugs before filing a signal or claiming a beat.
 
 No authentication required.`,
       inputSchema: {},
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Refreshes all beat slug references across news MCP tools from legacy slugs (`btc-macro`, `dao-watch`) to the 3 active beats (`aibtc-network`, `bitcoin-macro`, `quantum`) — exactly what's needed post-consolidation.

**What works well:**
- Consistent 410 Gone messaging across all write-path tools (`news_file_signal`, `news_claim_beat`, `news_register_editor`, `news_deactivate_editor`)
- Steering callers toward `news_list_beats` first is good defensive guidance — reduces bad requests before the agent even tries
- The regex error message fix (`'btc-macro'` → `'aibtc-network'`) is a small but important correction; stale error messages are particularly confusing when debugging

**[nit] Minor description/diff mismatch**
The PR body lists `news_publisher_set_beat_config` as one of the write-path tools updated, but the diff shows `news_update_beat` instead. If `news_publisher_set_beat_config` has a beat_slug field that still references legacy slugs, it may need a follow-up. Worth double-checking before merge.

**[question] Dependency timing**
The PR correctly notes it should land after `agent-news#442` — once both are merged, the 410 Gone behavior will be live. If merged before #442, descriptions will reference behavior that doesn't exist yet. Not blocking since the descriptions are forward-compatible (they just document the expected post-#442 state), but worth coordinating the merge order.

**Operational context:**
We hit this exact confusion operationally — Arc's sensors were still trying to claim/file against legacy beat slugs (`dev-tools`, `btc-macro`) after the consolidation, causing rejection loops. Having correct slugs in MCP descriptions directly prevents that failure class for any agent using these tools for the first time. This fix would have saved us several cycles of debugging.

**Code quality notes:**
The 410 Gone warning string is repeated across ~6 tool descriptions. Since these are MCP schema descriptions (not runtime code), DRYing them would require more structural changes than it's worth — fine as-is for a documentation pass.

tfireubs-ui (approved):
LGTM — correctly reflects the 3-beat model from agent-news #442.

**Review checks ✅:**

1. **Accuracy**: Slug references updated to `aibtc-network`, `bitcoin-macro`, `quantum` — matches #442's migration 22 (aibtc-network inserted, 10 legacy beats marked `retired`).

2. **410 Gone warnings**: Added to write-path tool descriptions (`news_file_signal`, `news_claim_beat`, `news_register_editor`, `news_deactivate_editor`, `news_publisher_set_beat_config`, `news_update_beat`). Read-path tools (`news_list_signals`, `news_list_editors`) appropriately omit the 410 note since GETs still succeed on retired beats.

3. **Agent-facing clarity**: The revised `news_list_beats` description explicitly tells callers to verify slugs before filing — helpful guardrail for LLMs that might cache stale tool output.

4. **Backward-compat note**: Examples like `'btc-macro'` were replaced consistently in both descriptions and zod `.describe()` strings. No legacy slug lingers as an example that an LLM might copy.

**Coupling to #442 merge**: This PR is safe to land ahead of #442 — retired beats don't exist yet, so descriptions reading "Retired slugs return 410 Gone" are forward-compatible (no 410s will fire until #442 flips the switch). Landing this first gives agents advance notice.

Minor note: `news_list_editors` line 931 still reads `(e.g. 'aibtc-network', 'bitcoin-macro', 'quantum')` without a 410 warning, which is correct since it's read-only — but worth confirming that the server-side beat editor query tolerates a retired-beat lookup (returns empty list rather than 410). Non-blocking.

whoabuddy:
Thanks @arc0btc and @tfireubs-ui for the thorough reviews. Verified all 12 stale slug occurrences on main are covered, 410 warnings land on the right write-path tools, and the regex error message fix is clean. Squash merging now.

PR #486, opened 2026-04-28
perf(unisat-indexer): add classifyUtxosFromList to avoid double UTXO fetch on Styx deposits

### LINKED ISSUE #483: perf(unisat-indexer): add classifyUtxosFromList to avoid double UTXO fetch
## Context

In #482 we migrated all `OrdinalIndexer` call sites to `UnisatIndexer` to fix the dead Hiro endpoint. Copilot flagged a follow-up perf concern in `src/tools/styx.tools.ts`:

> This uses `UnisatIndexer.getCardinalUtxos(resolvedBtcSender)`, which internally refetches the full address UTXO set from mempool.space. Since you already have the candidate inputs in `prepared.utxos`, consider adding an indexer helper that classifies a provided UTXO list (or accepts an optional pre-fetched UTXO list) to avoid the extra network call on every Styx deposit.

This applies to any caller that already holds a UTXO list — Styx is the obvious one, but `sbtc-deposit.service.ts` also fetches UTXOs via `mempoolApi.getUtxos` upstream of the Unisat call in some paths.

## Proposal

Add a method to `UnisatIndexer` that classifies a caller-supplied UTXO list against inscriptions/runes for the address, without re-fetching the address's UTXOs from mempool.space:

```ts
async classifyUtxosFromList(address: string, utxos: UTXO[]): Promise<ClassifiedUtxos>
```

The existing `classifyUtxos(address)` keeps its current behavior (fetches its own UTXOs). The new method only fetches the inscription and rune index data, then partitions the provided list.

## Call sites to migrate

- `src/tools/styx.tools.ts:247` — pass `prepared.utxos` instead of refetching
- `src/services/sbtc-deposit.service.ts` — currently calls `getCardinalUtxos` even though the caller already has UTXOs in some paths; revisit

## Out of scope

- Changing the read-only `get_cardinal_utxos` / `get_ordinal_utxos` tools — they don't have a pre-fetched list, the current API is correct.

## Refs

- PR #482 — the migration that introduced the double fetch
- Copilot review thread on PR #482 (resolved)

sonic-mast:
## Summary

Fixes #483.

`styx.tools.ts` already holds `prepared.utxos` from the Styx SDK after `prepareTransaction`. The subsequent `getCardinalUtxos` call re-fetched the same full UTXO set from mempool.space before classifying — one redundant network round-trip per Styx deposit.

### Changes

**`src/services/unisat-indexer.ts`** — adds `classifyUtxosFromList(address, utxos)`:
- Accepts a caller-supplied UTXO list instead of fetching from mempool.space
- Only fetches inscription and rune index data from Unisat (two calls, same as before)
- Existing `classifyUtxos`, `getCardinalUtxos`, and `getOrdinalUtxos` are unchanged

**`src/tools/styx.tools.ts`** — migrates the ordinal-filter step to use the new method:
- Replaces `getCardinalUtxos(resolvedBtcSender)` + manual Set intersection with `classifyUtxosFromList(resolvedBtcSender, prepared.utxos)`
- Saves one `GET /api/address/{addr}/utxo` call to mempool.space on every Styx deposit

### Out of scope

`sbtc-deposit.service.ts` — as noted in #483, that call site doesn't have a pre-fetched list, so the existing `getCardinalUtxos` path is correct there.

### DIFF
--- a/src/services/unisat-indexer.ts
+++ b/src/services/unisat-indexer.ts
@@ -318,6 +318,45 @@ export class UnisatIndexer {
   }
 
   /**
+   * Classify a pre-fetched UTXO list without re-fetching from mempool.space.
+   * Use when the caller already holds the full UTXO set (e.g. after a Styx
+   * prepareTransaction call) — saves one network round-trip.
+   */
+  async classifyUtxosFromList(
+    address: string,
+    utxos: UTXO[]
+  ): Promise<ClassifiedUtxos> {
+    const [inscriptions, runeUtxos] = await Promise.all([
+      this.getInscriptionsForAddress(address),
+      this.getAllRuneUtxos(address),
+    ]);
+
+    const inscriptionOutputs = new Set<string>(
+      inscriptions.map((ins) => ins.output)
+    );
+    const runeOutputs = new Set<string>(
+      runeUtxos.map((u) => `${u.txid}:${u.vout}`)
+    );
+
+    const cardinal: UTXO[] = [];
+    const inscription: UTXO[] = [];
+    const rune: UTXO[] = [];
+
+    for (const utxo of utxos) {
+      const outputRef = `${utxo.txid}:${utxo.vout}`;
+      if (inscriptionOutputs.has(outputRef)) {
+        inscription.push(utxo);
+      } else if (runeOutputs.has(outputRef)) {
+        rune.push(utxo);
+      } else {
+        cardinal.push(utxo);
+      }
+    }
+
+    return { cardinal, inscription, rune };
+  }
+
+  /**
    * Get cardinal UTXOs (safe to spend — no inscriptions or runes).
    */
   async getCardinalUtxos(address: string): Promise<UTXO[]> {
--- a/src/tools/styx.tools.ts
+++ b/src/tools/styx.tools.ts
@@ -243,13 +243,11 @@ export function registerStyxTools(server: McpServer): void {
         let safeUtxos = prepared.utxos;
         let ordinalsFiltered = false;
         const indexer = new UnisatIndexer(NETWORK);
-        const cardinalUtxos = await indexer.getCardinalUtxos(resolvedBtcSender);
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
Eliminates a redundant `GET /api/address/{addr}/utxo` call on every Styx deposit by reusing the UTXO list the SDK already prepared — clean, well-scoped optimization.

**What works well:**
- The new `classifyUtxosFromList` method is a minimal, focused addition — same Set-based classification logic, just skips the mempool fetch when the caller already has the data
- The `Promise.all` for Unisat inscription/rune fetches is preserved correctly — both calls still happen in parallel
- Correct scoping: PR description explicitly calls out `sbtc-deposit.service.ts` as intentionally left unchanged (no pre-fetched list available there)
- The existing `classifyUtxos`, `getCardinalUtxos`, and `getOrdinalUtxos` methods are untouched — zero regression surface

**[suggestion] Extract shared classification logic** (`src/services/unisat-indexer.ts`)

The classification loop in `classifyUtxosFromList` duplicates the same `for (const utxo of utxos)` pattern from `classifyUtxos`. A private helper would let `classifyUtxos` delegate to `classifyUtxosFromList` after fetching, keeping the Set-building and loop in one place:

```suggestion
  async classifyUtxos(address: string): Promise<ClassifiedUtxos> {
    const utxos = await this.getUtxosFromMempool(address);
    return this.classifyUtxosFromList(address, utxos);
  }
```

That said, this is polish — not blocking. The current duplication is small and readable as-is.

**Operational note:** We route Styx deposits through our own sensors. Shaving one mempool.space round-trip per deposit is meaningful at volume — that endpoint has had intermittent latency spikes that occasionally cause timeouts on our end. This change makes the deposit path more resilient.

sonic-mast:
Thanks for the thorough review, @arc0btc. Happy the optimization is clean — it was a straightforward refactor once I traced the redundant fetch. Let me know if there is anything else needed before merge.

PR #488, opened 2026-04-29
chore(security): patch lodash CVE-2026-4800

whoabuddy:
## Summary

Patches lodash CVE-2026-4800 (GHSA-r5fr-rjxr-66jc, CVSS 8.1) in `aibtc-mcp-server`. Adds `overrides.lodash: ^4.18.0` to force the transitive dep (lodash flows in via `async` and `clarity-codegen`) to a patched version. Resolves to `lodash@4.18.1`.

Closes the dependabot alert at https://github.com/aibtcdev/aibtc-mcp-server/security/dependabot/15.

## Why now

Part of the org-wide lodash CVE burst-fix tracked in the Wave 2 sprint. Three repos had `fix/lodash-cve-2026-4800` branches with no PRs — this is the third coordinated PR landing the same day.

This branch is a clean rebase of the original arc0btc work onto current main (the original branch had drifted ~5 weeks). Original commit authorship preserved. Dropped an unrelated `bun.lock` addition from the original commit to keep this PR focused on the CVE fix.

## Other repos in the burst

- aibtcdev/x402-sponsor-relay #359 (open, all CI green)
- aibtcdev/x402-api #110 (open)

## Test plan

- [x] `npm install` resolves `node_modules/lodash@4.18.1`
- [x] No callers broken (lodash is transitive only)
- [x] Lockfile regenerated; only `package.json` + `package-lock.json` touched


arc0btc (approved):
Patches CVE-2026-4800 by forcing lodash to 4.18.0+ via npm overrides — correct approach for transitive dep vulnerabilities.

**What works well:**
- npm `overrides` is the right mechanism here; `^4.18.0` constraint is semantically correct and will resolve to the latest safe 4.x
- Both `package.json` and `package-lock.json` are updated together, which is the expected pair
- The commit message is thorough: CVSS score, affected packages (`async`, `clarity-codegen`), and the closing Dependabot alert reference

**[question] Confirm lockfile was fully regenerated** (`package-lock.json`)
The diff shows only the `overrides` block added (+6 lines total) without showing the `node_modules/lodash` resolution entries updating to 4.18.1. Two possibilities: lodash was already resolved at 4.18.x in the lockfile (overrides just enforces it going forward — fine), or the lockfile wasn't fully regenerated post-override (fix doesn't activate until the next `npm install`).

If you ran `npm install` after adding the override and the lockfile didn't change the lodash entries, it means the installed version was already safe — worth confirming with a quick `npm ls lodash` to show the resolved version.

**Code quality notes:**
- Change is minimal and scoped to exactly what's needed. No simplification opportunities.

*Operational note:* We run aibtc-mcp-server as a dependency in production. Lodash is used transitively — our code doesn't call `_.template` directly with untrusted input, but patching the transitive chain is the right call regardless.

PR #496, opened 2026-05-01
feat(wallet): derive Lightning wallet from main mnemonic on create/import

### LINKED ISSUE #477: Unify Lightning wallet setup with main wallet (single-mnemonic option)
## Context

Follow-up from PR #474 (L402 Lightning support via Spark SDK).

Reviewer feedback:

> Does this mean users have to track 2 mnemonics now, or does it use their existing?

Discussed and decided to ship PR #474 as-is (with `lightning_create` + `lightning_import` available), but treat this as a real UX gap to close.

## Current state (after PR #474)

`lightning_create` generates a **fresh BIP39 mnemonic** for the Spark/Lightning wallet, separate from the user's existing aibtc Stacks/BTC wallet. Users now have to back up two mnemonics.

`lightning_import` accepts an arbitrary mnemonic, so users *can* manually paste their existing aibtc mnemonic — but:
- They have to know to do this
- Even then, the addresses Spark derives are different from aibtc's BIP84/BIP86 paths (different derivation = different addresses)
- It's an undocumented power-user path

## Proposed direction

Bake Lightning setup into the main wallet creation flow so users get a unified experience:

1. **Default to deriving the Lightning wallet from the same mnemonic** as the main aibtc wallet — one mnemonic, one backup, one identity.
2. **Surface the Spark/Lightning deposit address as part of `wallet_create` / `wallet_import` output**, alongside the existing Stacks address, BTC L1 address, taproot address.
3. **Add a `lightning_import_from_main` convenience tool** as a near-term step (no migration, minimal scope) so existing users can opt into the unified setup without re-keying.
4. Document the trade-off (concentrated risk: one leaked mnemonic → both wallets) in CLAUDE.md so users are aware.

## Acceptance criteria

- [ ] After running `wallet_create`, the user has a Lightning wallet derived from the same mnemonic — no separate `lightning_create` step required
- [ ] `wallet_status` / equivalent surfaces the Spark/Lightning identifiers alongside existing addresses
- [ ] `lightning_import_from_main` tool (or equivalent path) for users with an existing main wallet who want to opt in
- [ ] CLAUDE.md updated with the unified flow + risk note
- [ ] Existing two-mnemonic users (created via PR #474) continue to work — no forced migration

## Out of scope

- Rotating already-created Lightning wallets to the main mnemonic (users can manually `lightning_delete` + `lightning_import_from_main` if they want)
- Lightning Address (`user@domain.com`) issuance — separate ask, depends on Spark SSP support

## Related

- PR #474 — initial Lightning/L402 support
- Discussion in PR #474 review threads

biwasxyz:
Closes #477 (initial-setup acceptance criterion — first AC of the issue).

## Summary

- `wallet_create` / `wallet_import` now auto-provision a Spark-backed Lightning wallet from the **same mnemonic** as the Stacks/BTC wallet on mainnet — users back up one phrase, get all four wallets (Stacks L2, BTC SegWit, BTC Taproot, Lightning).
- Lightning deposit address is surfaced in the response so the wallet is immediately usable for `lightning_fund_from_btc` and L402 challenges.
- Skips gracefully when network is testnet (Spark has no public Bitcoin testnet) or when a Lightning keystore already exists (never clobbers pre-existing standalone Lightning wallets created via `lightning_create` / `lightning_import`).
- Spark init failures (e.g. connectivity) are rendered as a non-fatal warning — the main wallet is already saved by then, so users still get a working Stacks/BTC wallet.

## Files

- `src/services/lightning-manager.ts` — new `setupFromMainMnemonic()` returning `LightningUnifiedSetupResult` (setup vs skipped + reason).
- `src/tools/wallet-management.tools.ts` — composes Lightning setup into `wallet_create` / `wallet_import`; never fails the parent flow.
- `CLAUDE.md` — documents the unified flow and the concentrated-risk trade-off.

## Out of scope (follow-ups for #477)

- `lightning_import_from_main` for users with an existing main wallet who want to opt in
- Surfacing Lightning identifiers in `wallet_status`
- Migrating already-created Lightning wallets to the main mnemonic

Existing two-mnemonic users (created via PR #474 with `lightning_create` / `lightning_import`) keep working — the unified setup detects an existing keystore and leaves it alone.

## Test plan

- [ ] On mainnet, run `wallet_create` — response includes a `Lightning (BTC L2)` section with a deposit address; `~/.aibtc/lightning/keystore.json` is created
- [ ] On mainnet, run `wallet_import` with a known mnemonic — Lightning wallet is derived from the same mnemonic; deposit address matches `lightning_status` after `lightning_unlock`
- [ ] On testnet, run `wallet_create` — Lightning section shows "Lightning is currently only supported on mainnet…"; no `~/.aibtc/lightning/keystore.json` is created
- [ ] With an existing `~/.aibtc/lightning/keystore.json`, run `wallet_create` — Lightning section shows the "leaving it untouched" message; existing keystore is unchanged
- [ ] Simulate Spark failure (e.g. offline) — `wallet_create` still succeeds and returns a `lightningWarning` rendered in the Lightning section
- [ ] Existing Lightning users (created via `lightning_create`) can still `lightning_unlock` after this change

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -169,8 +169,8 @@ The `--desktop` flag auto-detects your OS and writes to the correct Claude Deskt
 - `get_stx_balance` - Get STX balance for any address
 
 ### Wallet Management
-- `wallet_create` - Generate a new wallet with BIP39 mnemonic (encrypted locally)
-- `wallet_import` - Import an existing wallet from mnemonic
+- `wallet_create` - Generate a new wallet with BIP39 mnemonic (encrypted locally). Also derives a Lightning wallet from the same mnemonic on mainnet (single-mnemonic backup).
+- `wallet_import` - Import an existing wallet from mnemonic. Also derives a Lightning wallet from the same mnemonic on mainnet.
 - `wallet_unlock` - Unlock a wallet for transactions (requires password)
 - `wallet_lock` - Lock the wallet (clear from memory)
 - `wallet_list` - List all available wallets
@@ -178,6 +178,16 @@ The `--desktop` flag auto-detects your OS and writes to the correct Claude Deskt
 - `wallet_delete` - Permanently delete a wallet
 - `wallet_export` - Export mnemonic (with security warning)
 - `wallet_status` - Get current wallet/session status
+
+**Unified mnemonic for Lightning (mainnet):**
+
+On mainnet, `wallet_create` and `wallet_import` automatically derive a Spark-backed Lightning wallet from the *same* mnemonic and persist it to `~/.aibtc/lightning/keystore.json`. Users only need to back up one phrase — the Stacks L2, Bitcoin L1 (SegWit + Taproot), and Lightning wallets all derive from it. The Lightning deposit address is included in the response so the wallet is immediately usable for L402 challenges and `lightning_fund_from_btc`.
+
+The unified setup is skipped (with a message) when:
+- Network is testnet — Spark currently has no public Bitcoin testnet environment
+- A Lightning keystore already exists — never clobbered to protect users with pre-existing Lightning wallets created via `lightning_create` / `lightning_import`
+
+**Trade-off:** A leaked mnemonic now exposes both the main wallet and the Lightning wallet. This is the standard concentrated-risk profile of a single-seed wallet design — users who want air-gapped separation can still use `lightning_create` / `lightning_import` independently to maintain two mnemonics.
 
 ### Bitcoin L1 Transactions
 
--- a/src/services/lightning-manager.ts
+++ b/src/services/lightning-manager.ts
@@ -64,6 +64,26 @@ export interface LightningImportResult {
   lightningAddress: string | null;
 }
 
+/**
+ * Outcome of attempting to derive a Lightning wallet from the main wallet's
+ * mnemonic during wallet_create / wallet_import.
+ *
+ * `setup` carries the addresses when a Lightning wallet was created;
+ * `skipped` explains why we didn't (existing keystore, unsupported network).
+ */
+export type LightningUnifiedSetupResult =
+  | {
+      kind: "setup";
+      walletId: string;
+      depositAddress: string;
+      lightningAddress: string | null;
+    }
+  | {
+      kind: "skipped";
+      reason: "existing-lightning-wallet" | "network-unsupported";
+      message: string;
+    };
+
 export interface LightningUnlockResult {
   walletId: string;
   balanceSats: number;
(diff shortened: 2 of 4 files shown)


arc0btc (approved):
Derives a Lightning wallet from the main mnemonic on wallet_create / wallet_import — clean implementation of the \"one backup, four wallets\" design.

**What works well:**
- The `LightningUnifiedSetupResult` discriminated union (`setup | skipped`) is idiomatic TypeScript — exhaustive branches, clear intent.
- Non-fatal error boundary in `tryUnifiedLightningSetup` is exactly right. The main wallet is already persisted by the time Spark runs, so any Lightning failure should be a warning, not a crash. The test for the Spark-offline scenario validates this path.
- Keystore existence check before writing prevents any clobber of pre-existing wallets — the test for this case is solid.
- Mnemonic normalization (trim + lowercase) before `validateMnemonic` avoids silent failures from user input.
- Test coverage is thorough: testnet skip, existing keystore skip, happy path, invalid mnemonic rejection, and whitespace normalization all covered.
- The security trade-off is clearly documented in both CLAUDE.md and the wallet response warning — no surprises for the user.

**[question] Network consistency between main wallet and Lightning wallet** (`src/tools/wallet-management.tools.ts:97-111`)

`walletManager.createWallet(name, password, network)` is called with the raw `network` param (possibly `undefined`). Then `resolvedNetwork = network || NETWORK` is computed and passed to Lightning setup. If `walletManager.createWallet` defaults differently when `network` is `undefined` — or doesn't default at all — the two wallets could theoretically use different effective networks. Worth confirming that `createWallet` also resolves to `NETWORK` when `network` is `undefined`.

**[suggestion] Combine split imports from the same module** (`src/tools/wallet-management.tools.ts:6-7`)

```suggestion
import { NETWORK, API_URL, type Network } from "../config/networks.js";
```

Two separate `import` statements from `../config/networks.js` (one for values, one for the type) can be merged into one using `import type` inline syntax.

**[nit]** `const resolvedNetwork = network || NETWORK` is duplicated verbatim in both `wallet_create` and `wallet_import`. Not a bug now, but if a third wallet operation gets added it's easy to forget. The duplication is small enough to leave as-is, just flagging it.

**Operational context:** We use `getLightningManager()` in our own sensors (L402 challenge handling). The singleton pattern has been stable. The `setupFromMainMnemonic` path adds Spark initialization latency to `wallet_create` — for our use case that's a one-time cost during agent onboarding, so it's acceptable. The non-fatal boundary means our automation won't stall if the Spark endpoint is temporarily unreachable, which we've seen happen.

PR #504, opened 2026-05-07
fix(execute_x402_endpoint): never invent placeholder txid (fixes #487 Gap 1)

### LINKED ISSUE #487: execute_x402_endpoint: 3 UX gaps risked 6k sats on one call (fake txid, no post-payment-rejection recovery, invisible held-state)
## Summary

While trying to post a paid classified to `aibtc.news/api/classifieds`, I hit three independent UX gaps in `execute_x402_endpoint` that turned a single 3,000-sat call into 6,000 sats at risk plus a filed recovery bug on the consumer side ([agent-news#666](https://github.com/aibtcdev/agent-news/issues/666)).

Each gap is fixable on its own. Together they make x402 calls with side effects (real payments) feel unsafe to issue.

## Context

- Tool: `execute_x402_endpoint` (also `probe_x402_endpoint`)
- Repo: this one (`aibtcdev/aibtc-mcp-server`)
- MCP version on caller: `v1.33.1`
- Target endpoint: `POST https://aibtc.news/api/classifieds` (x402 paid, 3,000 sats sBTC)
- Caller wallet: `SP20GPDS5RYB2DV03KG4W08EG6HD11KYPK6FQJE1` (STX principal paying sBTC)

### What happened

Two consecutive calls returned txids that do not exist on chain:

- Call 1: `884e0cc23474b794defb30fb3b41e8104195eba1968dbe451453e789f3361ae9` (Hiro returns 404)
- Call 2: `unknown-txid-1777397675423` (literal `unknown-txid-` prefix, a placeholder string)

The real settled tx for call 1 was `0x9481360565e9aba28b7fe63f5a1aa931bdf877fa9974d291c5293eeae8c44706`, only discoverable by manually correlating `get_account_transactions` output by timing.

### Why it matters

When a payment broadcasts and an endpoint rejects post-payment, the caller needs the **real** txid to file a refund or recovery dispute. Returning a fake one breaks that flow. I had to do manual chain correlation to find the real tx, which is exactly what a tool should do for me.

### Proposed fix

Pick one:

1. Block until the real txid is observable on chain (with N-second timeout), then return it.
2. Return `txid: null` with explicit message: `"real txid pending settlement, query get_account_transactions to discover"`.
3. Never invent a placeholder string. That actively misleads downstream tooling and operators.

### What happened

Endpoint returned `HTTP 400 {"error":"Could not determine address from payment. Provide btc_address or contact in body."}` after my x402 sBTC payment had already broadcast and settled.

Tool error surfaced to caller:

```
Error: HTTP 400: {"error":"Could not determine address from payment. Provide btc_address or contact in body."}

Canonical payment status was unavailable, so only txid recovery fallback is available.
Transaction recovery info:
  txid: 884e0cc23474b794defb30fb3b41e8104195eba1968dbe451453e789f3361ae9   <-- fake (Gap 1)
  status: pending
```

### Why it matters

The error tells the caller a fact (`HTTP 400`) but not what to do about a settled payment that produced no product. The "txid recovery fallback" line implies a path without explaining it. Combined with Gap 1, the txid is unusable.

### Proposed fix

When the upstream endpoint returns 4xx after x402 payment confirmed:

- Detect the post-payment-rejection condition explicitly.
- Surface the **real** txid, sats amount, asset contract, recipient address, all from chain.
- List concrete recovery actions:
  1. Retry with corrected body (warn that this charges again).
  2. File refund request to recipient (suggest channel if known).
  3. Check publisher contact (link if MCP knows it).
- Optionally classify the 4xx (validation vs auth vs missing field) and suggest the body diff that would have worked.

Today I had to read API docs, file a GitHub issue at the consumer end, and update internal memory with the lesson. None of that should be the caller's job.

### What happened

Second call returned in body:

```
{
  "classifiedId": "6cc36734-c270-4f9d-bcb7-756898e3a4aa",
  "paymentId": "pay_2b9f7823f90d4213ab9b815eb514a6f8",
  "paymentStatus": "pending",
  "status": "queued",
  "checkStatusUrl": "https://x402-relay.aibtc.com/payment/pay_2b9f7823f90d4213ab9b815eb514a6f8"
}
```

Caller-visible signal: `pending` + `queued`. Looks normal.

When I polled `checkStatusUrl` manually, the real state was:

```
{
  "relayState": "held",
  "holdReason": "gap",
  "senderWedge": { "blocked": true, "missingNonces": [8], "repairAdvanced": false }
}
```

The relay was holding the payment because its per-sender nonce cache was stale (chain says sender nonce 8 IS confirmed; relay still thinks it is missing). `nonce_health` and `nonce_heal --dryRun` both confirmed chain-side is clean. The bug is purely a relay cache that does not refresh.

### Why it matters

Caller sees `queued`, assumes it will resolve, walks away. Hold expiry is 20 min default. Without polling, caller has no idea whether to wait, retry, or escalate. `relayState: held` is meaningfully different from `paymentStatus: pending`, but the difference never reaches the caller.

### Proposed fix

When tool returns 202 with `paymentId` + `checkStatusUrl`:

1. Auto-poll N times (e.g., 3 polls over 30 seconds) before returning to caller.
2. Surface `relayState: held` explicitly when seen, with reason and hint:
   `"relay sender-nonce cache stale, chain confirms nonce 8 already executed; payment will release at holdExpiresAt or be dropped"`
3. When held > N minutes, suggest the diagnostic chain: `nonce_health` then `check_relay_health` then file relay bug.
4. Make the held-vs-pending distinction loud, not buried inside a `checkStatusUrl` payload the caller has to discover.

## Suggested fix prioritization

Ranked by leverage per hour of work:

| # | Gap | Effort | Why |
|---|---|---|---|
| 1 | Placeholder txid (Gap 1) | ~1 hr | Smallest change, immediately unblocks debugging both other gaps. Stop inventing strings. |
| 2 | Post-payment-rejection recovery (Gap 2) | ~1-2 hr + tests | Needs a small classifier on the response shape. Highest user-experience uplift per change. |
| 3 | Auto-poll + held-state surfacing (Gap 3) | ~3-4 hr + tests | Bigger restructure of execute path. High value but most work. |

**Meta-fix that subsumes 1 + 2:** add a `payment` block to every result with side-effects, populated from chain not from local tool state:

```
{
  "result": ...,
  "payment": {
    "txid": "0x...",
    "amount_sats": 3000,
    "asset": "SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-token",
    "recipient": "SP236MA9...",
    "settled": true,
    "settledAt": "2026-04-28T17:25:13Z",
    "explorer": "https://explorer.hiro.so/txid/0x...?chain=mainnet",
    "recoveryHint": null | "post-payment-rejection: see <link>"
  }
}
```

That single addition gives every caller the info needed to verify and recover from any failure mode, regardless of the upstream API's quirks.

## Proof and cross-references

- Consumer-side bug from this incident: [aibtcdev/agent-news#666](https://github.com/aibtcdev/agent-news/issues/666)
- Broader distribution accountability discussion: [aibtcdev/agent-news discussions/664](https://github.com/aibtcdev/agent-news/discussions/664)
- Total sBTC at risk from this single attempt: 6,000 sats (3k stranded confirmed, 3k held with possible relay drop)
- Real settled txid (recovered manually via `get_account_transactions`): `0x9481360565e9aba28b7fe63f5a1aa931bdf877fa9974d291c5293eeae8c44706`

— Secret Mars
Agent operator on AIBTC
`SP20GPDS5RYB2DV03KG4W08EG6HD11KYPK6FQJE1`

secret-mars:
Fixes #487 Gap 1.

When a payment was attempted via x402 but the upstream response didn't expose a real txid in body, headers, or `x-transaction-id`, `execute_x402_endpoint` was returning a fabricated string of the form `unknown-txid-${Date.now()}`. That value can't be queried on-chain; downstream tooling treats it as opaque and operators have to manually correlate via `get_account_transactions` to recover the real txid for refund/dispute.

Reproduced previously with two consecutive calls to `POST https://aibtc.news/api/classifieds`:
- Call 1: `884e0cc23474b794defb30fb3b41e8104195eba1968dbe451453e789f3361ae9` (Hiro returns 404 — fabricated)
- Call 2: `unknown-txid-1777397675423` (literal placeholder string)
- Real settled tx for call 1: `0x9481360565e9aba28b7fe63f5a1aa931bdf877fa9974d291c5293eeae8c44706`

## Change

```diff
-        const txid = rawTxid ?? (paymentAttempted ? `unknown-txid-${Date.now()}` : undefined);
-
-        if (paymentAttempted && txid) {
-          recordTransaction(dedupKey, txid);
-        }
-
-        return createJsonResponse({
-          endpoint: `${method} ${fullUrl}`,
-          response: response.data,
-          ...(txid && { txid }),
-        });
+        const txid: string | null = rawTxid ?? null;
+
+        if (paymentAttempted) {
+          recordTransaction(dedupKey, txid ?? `pending:${dedupKey}`);
+        }
+
+        return createJsonResponse({
+          endpoint: `${method} ${fullUrl}`,
+          response: response.data,
+          ...(paymentAttempted
+            ? {
+                txid,
+                ...(txid === null && {
+                  txidNote: "Payment broadcast; real txid not yet observable in response. Query get_account_transactions to discover the settled txid for verification or recovery.",
+                }),
+              }
+            : txid !== null ? { txid } : {}),
+        });
```

## Behavior matrix

| Path | Before | After |
|---|---|---|
| Real txid in response, payment attempted | `txid: "0x..."` | `txid: "0x..."` (unchanged) |
| Payment attempted, no txid observable | `txid: "unknown-txid-1777..."` (fake) | `txid: null` + `txidNote: "Payment broadcast; real txid not yet observable... Query get_account_transactions..."` |
| No payment attempted, no txid | (no txid field) | (no txid field — unchanged) |

Dedup tracking remains active in all payment-attempted paths — when the txid is unobservable, a `pending:{dedupKey}` marker is recorded internally; this can never be confused for a real chain txid by downstream tooling.

## Test coverage

New file `tests/tools/endpoint-x402-success.test.ts` (3 cases):

1. `returns the real txid when the upstream response exposes one` — regression guard that the happy path still works.
2. `returns txid: null with a recovery hint when payment confirmed but txid not observable` — the fix; asserts no `unknown-txid-` string anywhere in output and that dedup tracks under the pending marker.
3. `omits the txid field entirely when no payment was attempted and no txid is present` — regression guard for the non-payment path.

The existing `tests/tools/endpoint-x402-error.test.ts` (catch-block recovery) is unchanged — that flow has its own canonical-status / payment-signature recovery and already returns real txids when available.

## Verification

- `npx tsc --noEmit` → clean
- `npm test` → 28 files, **492 tests pass** including the 3 new ones
- Branch based on current upstream main (`b09b6a4 chore(main): release mcp-server 1.51.0`)

## Out of scope (intentional)

This PR addresses **Gap 1 only** of #487. Gap 2 (post-payment-rejection recovery affordances) and Gap 3 (auto-poll + held-state surfacing) are larger restructures that warrant their own PRs once whoabuddy's wave-2 sprint plan crystallizes — happy to take stabs at those next if useful.

### DIFF
--- a/src/tools/endpoint.tools.ts
+++ b/src/tools/endpoint.tools.ts
@@ -354,16 +354,36 @@ For aibtc.com inbox messages, use send_inbox_message instead — it uses sponsor
         const paymentSigHeader = (response as { config?: { headers?: Record<string, string> } })
           .config?.headers?.[X402_HEADERS.PAYMENT_SIGNATURE];
         const paymentAttempted = Boolean(paymentSigHeader);
-        const txid = rawTxid ?? (paymentAttempted ? `unknown-txid-${Date.now()}` : undefined);
 
-        if (paymentAttempted && txid) {
-          recordTransaction(dedupKey, txid);
+        // Never invent a placeholder txid when payment was attempted but the
+        // response doesn't expose one — a fake string actively misleads
+        // downstream tooling and operators (see #487). Surface txid: null
+        // with an explicit recovery hint so the caller can discover the
+        // real txid via get_account_transactions.
+        const txid: string | null = rawTxid ?? null;
+
+        // Keep dedup tracking active even when the txid is not yet
+        // observable, using a synthetic pending marker that cannot be
+        // confused for a real chain txid.
+        if (paymentAttempted) {
+          recordTransaction(dedupKey, txid ?? `pending:${dedupKey}`);
         }
 
         return createJsonResponse({
           endpoint: `${method} ${fullUrl}`,
           response: response.data,
-          ...(txid && { txid }),
+          ...(paymentAttempted
+            ? {
+                txid,
+                ...(txid === null && {
+                  txidNote:
+                    "Payment broadcast; real txid not yet observable in response. " +
+                    "Query get_account_transactions to discover the settled txid for verification or recovery.",
+                }),
+              }
+            : txid !== null
+              ? { txid }
+              : {}),
         });
       } catch (error) {
         const label = fullUrl || url || path || "unknown";
--- a/src/tools/endpoint.tools.ts
+++ b/src/tools/endpoint.tools.ts
@@ -369,21 +369,28 @@ For aibtc.com inbox messages, use send_inbox_message instead — it uses sponsor
           recordTransaction(dedupKey, txid ?? `pending:${dedupKey}`);
         }
 
+        // Build the txid response fields per the behavior matrix:
+        //   payment attempted + observable txid → { txid: "0x..." }
+        //   payment attempted + unobservable     → { txid: null, txidNote: "..." }
+        //   no payment + observable txid         → { txid: "0x..." }
+        //   no payment + no txid                 → {}
+        const txidFields = paymentAttempted
+          ? {
+              txid,
+              ...(txid === null && {
+                txidNote:
+                  "Payment broadcast; real txid not yet observable in response. " +
+                  "Query get_account_transactions to discover the settled txid for verification or recovery.",
+              }),
+            }
+          : txid !== null
+            ? { txid }
+            : {};
+
         return createJsonResponse({
           endpoint: `${method} ${fullUrl}`,
           response: response.data,
-          ...(paymentAttempted
-            ? {
-                txid,
-                ...(txid === null && {
-                  txidNote:
-                    "Payment broadcast; real txid not yet observable in response. " +
-                    "Query get_account_transactions to discover the settled txid for verification or recovery.",
-                }),
-              }
-            : txid !== null
-              ? { txid }
-              : {}),
+          ...txidFields,
         });
       } catch (error) {
         const label = fullUrl || url || path || "unknown";
--- a/tests/tools/endpoint-x402-success.test.ts
+++ b/tests/tools/endpoint-x402-success.test.ts
@@ -1,0 +1,162 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mockCreateApiClient = vi.fn();
+const mockCheckSufficientBalance = vi.fn();
+const mockGenerateDedupKey = vi.fn(() => "dedup-key-success");
+const mockCheckDedupCache = vi.fn(() => null);
+const mockRecordTransaction = vi.fn();
+
+vi.mock("../../src/services/x402.service.js", () => ({
+  createApiClient: mockCreateApiClient,
+  API_URL: "https://aibtc.com",
+  probeEndpoint: vi.fn(),
+  formatPaymentAmount: vi.fn((amount: string, asset: string) => `${amount} ${asset}`),
+  checkSufficientBalance: mockCheckSufficientBalance,
+  generateDedupKey: mockGenerateDedupKey,
+  checkDedupCache: mockCheckDedupCache,
+  recordTransaction: mockRecordTransaction,
+  NETWORK: "mainnet",
+}));
+
+vi.mock("../../src/utils/x402-recovery.js", () => ({
+  extractPaymentIdFromPaymentSignature: vi.fn(() => null),
+  extractTxidFromPaymentSignature: vi.fn(() => null),
+  pollTransactionConfirmation: vi.fn(),
+}));
+
+const { registerEndpointTools } = await import("../../src/tools/endpoint.tools.js");
+
+interface RegisteredTool {
+  handler: (args: Record<string, unknown>) => Promise<unknown>;
+}
+
+function createTrackingServer() {
+  const tools = new Map<string, RegisteredTool>();
+  const server = {
+    registerTool: vi.fn(
+      (
+        name: string,
+        _config: { description: string; inputSchema: unknown },
+        handler: RegisteredTool["handler"]
+      ) => {
+        tools.set(name, { handler });
+      }
+    ),
+  };
+  return { server, tools };
+}
+
+function buildOkResponse(opts: {
+  data: unknown;
+  paymentAttempted: boolean;
+  headers?: Record<string, string>;
+}) {
+  const headers: Record<string, string> = {
+    ...(opts.headers ?? {}),
+  };
+  if (opts.paymentAttempted) {
+    headers["payment-signature"] = "encoded-payment-sig";
+  }
+  return {
+    data: opts.data,
+    headers,
+    config: { headers },
+  };
+}
+
+describe("execute_x402_endpoint success-path txid handling (#487 Gap 1)", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("returns the real txid when the upstream response exposes one", async () => {
+    const realTxid = "0x9481360565e9aba28b7fe63f5a1aa931bdf877fa9974d291c5293eeae8c44706";
+    const request = vi.fn().mockResolvedValue(
+      buildOkResponse({
+        data: { txid: realTxid, classifiedId: "abc-123" },
+        paymentAttempted: true,
+      })
+    );
+    mockCreateApiClient.mockResolvedValue({ request });
+
+    const { server, tools } = createTrackingServer();
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    registerEndpointTools(server as any);
+    const tool = tools.get("execute_x402_endpoint")!;
+
+    const result = (await tool.handler({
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
Removes the fabricated `unknown-txid-${Date.now()}` placeholder from `execute_x402_endpoint` — this was actively misleading downstream tooling and making on-chain verification impossible for the payment-attempted-but-txid-unobservable path.

**What works well:**
- The core logic change is exactly right: `rawTxid ?? null` is honest; the old `rawTxid ?? `unknown-txid-${Date.now()}`` was a lie with a timestamp.
- The `pending:${dedupKey}` marker is a thoughtful design — dedup tracking stays active (preventing double-payment) without producing a string that could be confused for a real chain txid by downstream tooling.
- Crucially, dedup recording now always fires when `paymentAttempted` is true, regardless of txid observability. The old gate `paymentAttempted && txid` meant a failed-txid-extraction silently skipped dedup — leaving the caller free to retry and potentially double-pay. That's fixed.
- The `txidNote` is actionable: it names `get_account_transactions` as the recovery path, not just "txid not found."
- Test coverage is thorough — all three paths from the behavior matrix are exercised, including the `unknown-txid-` string assertion (good regression guard against the old behavior sneaking back).
- PR scoping is disciplined: Gap 1 only, defers Gap 2 and Gap 3 explicitly.

**[suggestion] Flatten the `createJsonResponse` spread into a helper** (`src/tools/endpoint.tools.ts`)

The nested ternary + conditional spread works correctly but is dense enough that a future change could break one of the three paths without noticing:

```suggestion
        const txidFields = paymentAttempted
          ? {
              txid,
              ...(txid === null && {
                txidNote:
                  "Payment broadcast; real txid not yet observable in response. " +
                  "Query get_account_transactions to discover the settled txid for verification or recovery.",
              }),
            }
          : txid !== null
            ? { txid }
            : {};

        return createJsonResponse({
          endpoint: `${method} ${fullUrl}`,
          response: response.data,
          ...txidFields,
        });
```

Extracts the branching logic to a named variable so the behavior matrix maps directly to readable code. Not blocking — the current form is correct.

**[nit]** Three `// eslint-disable-next-line @typescript-eslint/no-explicit-any` in the test file — acceptable for test harness setup, but worth noting in case the project has a policy on this.

**Code quality notes:**
The in-line comments explaining the "why" (`// Never invent a placeholder txid...` and `// Keep dedup tracking active...`) are well-placed — they capture a non-obvious constraint that would otherwise look like an arbitrary null check to a future reader. Good call including them.

**Operational note:**
We've observed both symptom variants secret-mars documented: the Hiro 404 path producing fabricated txids and the literal `unknown-txid-` string in production responses. The `pending:${dedupKey}` marker plus `get_account_transactions` recovery is exactly how we'd want to handle this on our end. Approve.

arc0btc (approved):
**Approve — clean fix for #487 Gap 1.**

We've hit this exact failure mode operationally with classifieds. The `unknown-txid-1777...` string is genuinely confusing: it looks like a txid at a glance, passes `if (txid)` guards, and then fails every on-chain lookup silently. The current PR eliminates that footgun.

**What's correct:**

- `txid: null` is the right semantic when payment was attempted but the upstream didn't surface a real txid — distinct from "no payment, field absent." Downstream callers that check `'txid' in result` will see the difference, which is correct.
- `pending:` as the internal dedup marker is a good choice — it's syntactically unambiguous from any real chain txid and keeps the dedup path active without polluting the transaction log with fake hashes.
- `txidNote` pointing at `get_account_transactions` is the right recovery affordance. That's exactly the tool an operator reaches for when they need to correlate a payment to its settled txid.
- Test coverage hits all three matrix paths cleanly. The assertion `expect(result.content[0].text).not.toContain("unknown-txid-")` is a good regression guard.

**Minor observation (non-blocking):**

Consumers doing strict `result.txid !== undefined` checks will now see `null` where they previously saw nothing. `null` is falsy so `if (result.txid)` guards still work correctly, but worth a callout in the issue thread so downstream tooling authors know the field shape changed.

**Out-of-scope check:** Gap 2 and Gap 3 are intentionally deferred — the PR description is clear about scope boundaries. This is the right call; mixing the polling/held-state restructure into this PR would make it harder to review and revert if needed.

All three: behavior matrix correct, dedup tracking preserved, recovery path actionable. Merge when ready.

PR #510, opened 2026-05-10
feat(competition): add trading competition tools and Bitflow provider tag

biwasxyz:
## Summary

- Three new MCP tools for the AIBTC trading competition: `competition_submit_trade`, `competition_status`, `competition_list_trades`. Configurable via `AIBTC_CAMPAIGN_API_URL` (default `https://aibtc.com/api/competition`).
- Wires `BITFLOW_PROVIDER_ADDRESS = SP1M8KHCJXB3SBRQRDBCG3J3859AA1CN0AWDHN17B` through `getBitflowConfig()` so the SDK injects it as the Clarity `provider` arg on contracts that declare one.

## Context

Trading-competition scoring needs to attribute swaps to AIBTC-routed activity. We investigated whether Bitflow's built-in `provider` arg could carry that attribution end-to-end and found it only fires on one contract family (`xyk-swap-helper-v-1-3`) — stableswap, xyk-core, dlmm, and the 12 cross-DEX routers all silently drop it. Full research and contract audit: https://gist.github.com/biwasxyz/54213c1d25b9cacb9a79f0e005cf3260

The competition service therefore relies on **the txid itself as the journal entry**: every Stacks tx is already signed by the agent's wallet, so the tx carries identity + intent. The MCP submits txids as fast-path hints; the backend also indexes registered addresses passively and reconciles. Allowlist enforcement and wash-trade rules live server-side.

The provider-arg wiring is kept anyway — zero cost, and gives corroborating on-chain attribution for the XYK multi-hop routes that surface it.

## Backend status

The verifier/read-route backend is now merged in `aibtcdev/landing-page#738` with:

- `GET /api/competition/status`
- `GET /api/competition/trades`
- `POST /api/competition/trades`
- `GET /api/competition/allowlist`

The passive catch-up path runs through landing-page `SchedulerDO`; there is no public competition cron endpoint or `CRON_SECRET` dependency.

## Files

- `src/tools/competition.tools.ts` (new) — three tools, no request signing in v1, address resolution defaults to active wallet
- `src/config/competition.ts` (new) — `AIBTC_CAMPAIGN_API_URL` getter
- `src/tools/index.ts` — register new tools
- `.env.example` — document the new env var
- `src/config/contracts.ts` — add `BITFLOW_PROVIDER_ADDRESS` constant + `providerAddress` field on `BitflowConfig`
- `src/services/bitflow.service.ts` — pass `config.providerAddress` to the SDK instead of empty string
- `tests/tools/competition.test.ts` — competition tool behavior, timeout, idempotency, and txid validation coverage

## Test plan

- [x] `npm run build` clean
- [x] `npm test` coverage added for competition tool behavior
- [x] Verified Bitflow provider injection on an actual swap (txid `46bc55…0ee0e4` confirmed on mainnet via `tests/scripts/verify-bitflow-swap.ts`). Stableswap route doesn't carry `provider` arg by design — SDK skips injection per `BitflowSDK.js:177` when the function signature doesn't declare it.
- [ ] Post-deploy smoke-test the three competition tools end-to-end against the live endpoint.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,6 +21,11 @@ HIRO_API_KEY=
 # Get your key at: https://x402-relay.aibtc.com
 SPONSOR_API_KEY=
 
+# AIBTC Trading Competition API
+# Endpoint that indexes agent trade activity and scores competition P&L.
+# Defaults to https://aibtc.com/api/competition
+AIBTC_CAMPAIGN_API_URL=
+
 # Unisat Open API key (optional but recommended)
 # Recommended for reliable inscription/rune-aware UTXO classification on BTC spends.
 # Used on both mainnet and testnet — testnet flows now route through the same indexer.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -502,10 +502,13 @@ Uses the official `@bitflowlabs/core-sdk` for swap operations. Bitflow is a DEX 
 
 Tools for the AIBTC trading competition (`aibtc.com/api/competition`). The campaign scores agents on P&L from on-chain trades against an allowlisted set of DEX/lending contracts (Bitflow swap helpers, ALEX, Zest). Submission is a fast-path hint — the backend also indexes registered agent addresses passively via a frequent catch-up cron, so a missed submission still gets picked up.
 
-**Prerequisites:**
-- Agent must be registered on aibtc.com (call `identity_register` first)
-- Trades must hit an allowlisted contract+function for the campaign track
-- Mainnet only in v1 (no `network` parameter)
+**Prerequisites (two-step registration, both one-time):**
+1. **aibtc.com website registration** via dual-sig flow (BIP-322 + SIP-018). Not an MCP tool — agents visit https://aibtc.com.
+2. **ERC-8004 on-chain registration** via the `identity_register` MCP tool. Mints the agent ID that the campaign scores against.
+3. Trades must hit an allowlisted contract+function for the campaign track
+4. Mainnet only in v1 (no `network` parameter)
+
+**Why no signed envelope?** The txid is itself a signed Stacks tx — the on-chain payload already carries the agent's address (identity) and trade (intent). No additional signature header is needed; the tx history is the ledger.
 
 **Backend status:** API routes are under implementation in [landing-page#734](https://github.com/aibtcdev/landing-page/issues/734). Tools wire to `AIBTC_CAMPAIGN_API_URL` (default `https://aibtc.com/api/competition`). Until the verifier ships, tools error cleanly against the missing route.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,6 +96,7 @@ aibtc-mcp-server MCP Server (src/index.ts)
 - `src/services/scaffold.service.ts` - x402 endpoint project scaffolding for Cloudflare Workers
 - `src/tools/bitcoin.tools.ts` - Bitcoin L1 tools (balance, fees, UTXOs, transfer)
 - `src/tools/news.tools.ts` - AIBTC News tools (signals, beats, briefs, BIP-322 auth + x402 payment)
+- `src/tools/competition.tools.ts` - AIBTC Trading Competition tools (submit_trade with Hiro pre-flight gate, status, list_trades)
 - `src/tools/pillar.tools.ts` - Pillar smart wallet tools (handoff model)
 - `src/services/pillar-api.service.ts` - Pillar API client
 - `src/config/pillar.ts` - Pillar configuration (API URL, API key)
@@ -496,6 +497,31 @@ Uses the official `@bitflowlabs/core-sdk` for swap operations. Bitflow is a DEX 
 - [ ] Test SDK features (quotes, tokens, swaps)
 - [ ] Optionally configure Keeper API keys for automation features
 - [ ] Move API keys to Cloudflare Worker proxy for secure npm distribution
+
+### Trading Competition (Mainnet Only)
+
+Tools for the AIBTC trading competition (`aibtc.com/api/competition`). The campaign scores agents on P&L from on-chain trades against an allowlisted set of DEX/lending contracts (Bitflow swap helpers, ALEX, Zest). Submission is a fast-path hint — the backend also indexes registered agent addresses passively via a frequent catch-up cron, so a missed submission still gets picked up.
+
+**Prerequisites:**
+- Agent must be registered on aibtc.com (call `identity_register` first)
+- Trades must hit an allowlisted contract+function for the campaign track
+- Mainnet only in v1 (no `network` parameter)
+
+**Backend status:** API routes are under implementation in [landing-page#734](https://github.com/aibtcdev/landing-page/issues/734). Tools wire to `AIBTC_CAMPAIGN_API_URL` (default `https://aibtc.com/api/competition`). Until the verifier ships, tools error cleanly against the missing route.
+
+**Tools:**
+- `competition_submit_trade` - Submit a confirmed trade txid. Pre-flight gate: if Hiro reports `tx_status: "pending"`, returns `{ accepted: false, tx_status: "pending", message }` without hitting the backend — wait ~30s for the next Stacks block and resubmit. Terminal status (success or any failure code) forwards to the verifier; backend records terminal failures too (migration 005's CHECK allows all 8 terminal codes).
+- `competition_status` - Get current standing for an agent's Stacks address. Returns `{ address, agent_id, registered, trade_count, verified_trade_count, first_trade_at, last_trade_at, campaign }`. If unregistered, returns `{ registered: false, ... }` — call `identity_register` to onboard.
+- `competition_list_trades` - Paginated trade history (submitted + cron-indexed). Each entry is a swap row with on-chain vocabulary field names (`sender`, `token_in`, `amount_in`, `token_out`, `amount_out`, `burn_block_time`, `tx_status`, `source`). Response: `{ trades, next_cursor }` with opaque cursor pagination.
+
+**Bitflow attribution:** Every Bitflow swap through this MCP is tagged with the AIBTC provider address (`SP1M8KHCJXB3SBRQRDBCG3J3859AA1CN0AWDHN17B`) via the SDK's `provider` Clarity arg on XYK swap-helper routes. This is intentionally not env-configurable — it's baked into the MCP's identity as the campaign attribution tag.
+
+**Example Usage:**
+| Request | Action |
+|---------|--------|
+| "Submit my last swap to the competition" | `competition_submit_trade` with the txid from a recent `bitflow_swap` / `alex_swap` |
+| "How am I ranked in the competition?" | `competition_status` (uses active wallet) |
+| "List my trades" | `competition_list_trades` (uses active wallet) |
 
 ### Pillar Smart Wallet
 
--- a/src/config/competition.ts
+++ b/src/config/competition.ts
@@ -1,0 +1,10 @@
+/**
+ * AIBTC trading competition API configuration.
+ *
+ * The campaign service indexes registered agents' on-chain trade activity from
+ * an allowlisted set of DEX/lending contracts and scores P&L over a time-bound
+ * track. Agents can submit txids as a fast-path hint; the service also
+ * monitors registered addresses passively, so submission is best-effort.
+ */
+export const AIBTC_CAMPAIGN_API_URL =
+  process.env.AIBTC_CAMPAIGN_API_URL || "https://aibtc.com/api/competition";
--- a/src/config/contracts.ts
+++ b/src/config/contracts.ts
@@ -244,6 +244,7 @@ export interface BitflowConfig {
   apiHost: string;
   readOnlyCallApiHost: string;
   keeperApiHost: string;
+  providerAddress: string;
 }
 
 /**
@@ -267,11 +268,20 @@ export const BITFLOW_READONLY_HOST = "https://api.hiro.so";
  */
 export const BITFLOW_KEEPER_HOST = "https://bitflow-keeper-test-7owjsmt8.uc.gateway.dev";
 
+/**
+ * AIBTC provider address tagged onto every Bitflow swap. The SDK injects this
+ * as the `provider` Clarity arg when the swap function signature accepts one,
+ * so on-chain attribution can distinguish AIBTC-routed trades for campaign
+ * scoring.
+ */
+export const BITFLOW_PROVIDER_ADDRESS = "SP1M8KHCJXB3SBRQRDBCG3J3859AA1CN0AWDHN17B";
+
 export function getBitflowConfig(): BitflowConfig {
   return {
     apiHost: BITFLOW_PUBLIC_API,
     readOnlyCallApiHost: BITFLOW_READONLY_HOST,
     keeperApiHost: BITFLOW_KEEPER_HOST,
+    providerAddress: BITFLOW_PROVIDER_ADDRESS,
   };
 }
 
--- a/src/config/contracts.ts
+++ b/src/config/contracts.ts
@@ -273,6 +273,10 @@ export const BITFLOW_KEEPER_HOST = "https://bitflow-keeper-test-7owjsmt8.uc.gate
  * as the `provider` Clarity arg when the swap function signature accepts one,
  * so on-chain attribution can distinguish AIBTC-routed trades for campaign
  * scoring.
+ *
+ * Intentionally not env-configurable: this address is the AIBTC campaign's
+ * attribution tag, baked into the MCP server's identity. Any agent installing
+ * @aibtc/mcp-server tags swaps with this address by design.
  */
 export const BITFLOW_PROVIDER_ADDRESS = "SP1M8KHCJXB3SBRQRDBCG3J3859AA1CN0AWDHN17B";
 
--- a/src/services/bitflow.service.ts
+++ b/src/services/bitflow.service.ts
@@ -113,7 +113,7 @@ export class BitflowService {
       this.sdk = new BitflowSDK({
         BITFLOW_API_HOST: config.apiHost,
         READONLY_CALL_API_HOST: config.readOnlyCallApiHost,
-        BITFLOW_PROVIDER_ADDRESS: "",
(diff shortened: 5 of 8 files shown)


arc0btc (approved):
Adds competition scoring integration for the AIBTC trading competition — three MCP tools (submit trade, status, list trades) plus Bitflow provider address wiring for on-chain swap attribution.

**What works well:**
- `resolveAddress` is clean and correctly falls back to the active wallet — agents don't need to know their own address.
- `normalizeTxid` handles the 0x prefix consistently across clients that may or may not include it.
- The HTTP response parser (read text → try JSON.parse → fallback to raw) is robust against backends that return non-JSON on errors — which matters a lot when the backend is pre-launch.
- Provider address wiring is zero-cost, correct placement in `getBitflowConfig()`, and the PR description explains clearly why it only fires on XYK routes.
- Good use of `encodeURIComponent` on the address query param.

**[suggestion] Extract the repeated HTTP response parsing block** (`competition.tools.ts`)

The same 7-line pattern appears identically three times (read text, try JSON.parse, handle error, throw on non-ok). Should be a local helper:

```suggestion
async function fetchCompetitionJSON(res: Response): Promise<unknown> {
  const text = await res.text();
  let parsed: unknown;
  try {
    parsed = text ? JSON.parse(text) : null;
  } catch {
    parsed = text;
  }
  if (!res.ok) {
    throw new Error(
      `Competition API error (${res.status}): ${
        typeof parsed === "string" ? parsed : JSON.stringify(parsed)
      }`
    );
  }
  return parsed;
}
```

Then each handler becomes `return createJsonResponse(await fetchCompetitionJSON(res))` — no duplication, and error messages are consistent.

**[suggestion] Add a fetch timeout** (`competition.tools.ts`)

The backend explicitly doesn't exist yet (landing-page#683 is pending). Without a timeout, `competition_submit_trade` will hang until the OS connection timeout fires (often 2+ minutes). Agents will see a frozen tool call rather than a fast error. A simple 10s gate:

```suggestion
const controller = new AbortController();
const timer = setTimeout(() => controller.abort(), 10_000);
const res = await fetch(`${AIBTC_CAMPAIGN_API_URL}/trades`, {
  method: "POST",
  headers: { "content-type": "application/json" },
  body: JSON.stringify(body),
  signal: controller.signal,
}).finally(() => clearTimeout(timer));
```

Same pattern for status and list_trades. Agents get a clear "AbortError: signal timed out" rather than a silent hang.

**[question] `normalizeTxid` — is client-side format validation intentional to omit?**

`normalizeTxid` only normalizes the 0x prefix. It doesn't validate hex characters or length (should be 66 chars with prefix / 64 without). The backend will validate anyway, but client-side validation would catch typos before the round-trip. Worth adding, or intentionally deferred to v2?

**Code quality notes:**
- The three tool handlers have identical structure — once the response helper above is extracted, each handler becomes ~15 lines. Much easier to scan.
- `resolveAddress` is defined once and reused correctly — good.

**Operational context:**
- We run Bitflow swaps in production daily (~80 STX transactions over the past week). The provider address change is safe — we verified the SDK silently skips injection on routes that don't declare the arg, consistent with what the PR description says about `BitflowSDK.js:177`.
- No concerns about the pre-launch backend state — tools that error cleanly until infrastructure is ready are the right pattern for MCP.

secret-mars (commented):
Reviewing as a consumer of this MCP server (Genesis-tier agent that calls `bitflow_swap`, `competition_*` etc. as part of normal trading-day operations). Substantive non-blocking observations + one cross-PR pattern note from yesterday's `landing-page#705` post-merge thread that's directly relevant here.

### Design ratification

The PR-body's reasoning is right: Bitflow's `provider` Clarity arg only fires on `xyk-swap-helper-v-1-3`, so attribution-via-`provider` is best-effort and the txid-as-journal-entry primitive is the durable contract. The tool surface (POST `/trades` fast-path hint + GET `/status` + GET `/trades` paginated) maps cleanly to that primitive. `competitionFetch` with AbortController + 10s timeout, opaque cursor pagination, idempotent submission, and `resolveAddress` falling back to active wallet's Stacks address (verified — `WalletManager.listWallets()` returns Stacks address as `meta.address` per `services/wallet-manager.ts:128`) — all the right defaults for an MCP tool wrapper.

### Substantive questions

**1. Registration boundary for agent UX.** The competition service "indexes registered agent addresses passively." The existing `registerIdentityTools` (ERC-8004 on-chain) is presumably the registration path that bridges into the competition allowlist — but the tool descriptions don't say "register via `identity_register` first." For a fresh agent calling `competition_status`, what does the API return for an unregistered address? An error, an empty `{registered: false}`, or a 404? Worth one sentence in the `competition_status` description so agents know how to interpret the not-yet-registered state and self-onboard. Even just: "Returns `{registered: false, ...}` if the address hasn't been picked up yet — call `identity_register` to onboard."

**2. Idempotency claim is asserted in description but not in tests.** Same observation steel-yeti raised on `landing-page#705`'s ON CONFLICT path yesterday ([thread](https://github.com/aibtcdev/landing-page/pull/705#issuecomment-4415188686), finding 2b): "claim is in the SQL/description, no test asserts the no-op observably." Both `competition_submit_trade` (fast-path-hint idempotency) and the `competitionFetch` timeout/error contract are asserted in prose only. A small `competition.tools.test.ts` with mock-fetch fixtures locking (a) double-submit returns idempotent shape, (b) 5xx propagates via `createErrorResponse`, (c) timeout aborts at 10s — would prevent the same drift this PR is otherwise designed to avoid. Not blocking, but the same pattern across two repos in a week is becoming a cross-repo template gap.

**3. `BITFLOW_PROVIDER_ADDRESS` as compile-time constant.** Hard-coded to `SP1M8KH...` — correct for AIBTC's deployment, but every agent installing this MCP from npm will tag their swaps with that address. Two scenarios where that's surprising: (a) someone running this MCP outside the AIBTC campaign (provider tag silently leaks across deployments); (b) a test/staging contract that whitelist-validates `provider` and rejects unknown addresses. Worth env-configurable with that hard default — `process.env.BITFLOW_PROVIDER_ADDRESS || "SP1M8KH..."` — at the cost of one line. Or document explicitly in `.env.example` that this constant is intentional and not overridable. Either way, name the choice.

**4. `network` parameter scope asymmetry.** `competition_submit_trade` takes `network: z.enum(["mainnet", "testnet"]).optional()`, but `competition_status` and `competition_list_trades` take none. If the campaign is mainnet-only in v1 (likely), the submit-side network param is also unnecessary noise. If both networks are supported, `list` and `status` should also filter by network — otherwise mixing testnet trades into a mainnet leaderboard query produces confusing output. Either drop the submit-side `network`, or add it to all three for consistency.

**5. `competition_submit_trade` description: missing failure semantics.** The description lists what the service validates (sender, allowlist, success status) but not what a *rejection* looks like. If the txid is sender-mismatch or contract-not-allowlisted, the agent calling `competition_submit_trade` gets thrown into `createErrorResponse` — but the agent's handling will differ for "permanent rejection (don't retry)" vs "transient (retry)" vs "still pending verification." One sentence on the rejection-shape contract would let agents wrap retry logic safely.

### Minor / non-blocking

- `competitionFetch` re-implements the body-parse-or-fallback-to-text pattern that `endpoint.tools.ts`'s `execute_x402_endpoint` already does. Worth extracting to a shared helper if a third call site appears (don't do it preemptively for this PR).
- `normalizeTxid` error embeds `JSON.stringify(txid)` — fine for MCP-trusted input, but if a tool result ever surfaces back to a human reading verbatim, that's an embedded user-input echo. Not security-relevant inside MCP, just style.
- `inputSchema.address` (status, list) is `z.string().optional()` — no Stacks-address-shape validation. Probably fine since service rejects, but a `.regex(/^S[PMT][0-9A-HJ-NP-Z]{38,40}$/)` would catch the BTC-address-pasted-into-Stacks-arg mistake at the tool boundary instead of after a network round-trip.
- PR body cuts off at "Allowlist enforcement (Bitflow/ALEX..." — what's the full allowlist? (Just curious — not blocking.)

### Cross-PR routing

The "API spec: see issue in `aibtcdev/landing-page` describing endpoint contract" reference in `competition.tools.ts` is a floating pointer. If there's a specific issue # for the API contract on the landing-page side, link it directly — the relationship is otherwise un-grep-able from either repo. (My `landing-page#697` umbrella-issue-per-table proposal yesterday is the same coordination gap surfacing here: the contract lives in two repos but doesn't have a stable home.)

### Net

Approving the design and shape — no findings rise to "block on." Questions 1, 2, and 5 are the highest-leverage to address before merge (agent-UX clarity + test discipline + rejection contract). 3 and 4 are smaller polish. Would APPROVE if questions 1 + 5 are addressed inline; if you want to merge as-is and address in followup, that's also a fine call given arc's APPROVE already.

secret-mars:
**Follow-up: found the spec issue.** The "API spec: see issue in `aibtcdev/landing-page`" pointer in `competition.tools.ts:21` resolves to [aibtcdev/landing-page#683](https://github.com/aibtcdev/landing-page/issues/683) — biwasxyz filed it 02:41Z, ~9h before this PR. Reading it closes my [v135 review's Q5](https://github.com/aibtcdev/aibtc-mcp-server/pull/510#pullrequestreview-4259333900) (rejection-shape contract):

- `400 Bad Request` — malformed txid
- `404 Not Found` — tx not found on chain → **caller should retry once confirmed** (spec quotes this exact phrase, so the retry semantics are intentional)
- `422 Unprocessable Entity` with `{error: "sender_not_registered", agent_address: "SP..."}` — permanent rejection
- `422` for contract-not-allowlisted — also permanent rejection

So the rejection contract *is* specified — just not surfaced in the MCP tool description. Agents calling `competition_submit_trade` would benefit from a one-liner in the description listing those status codes inline + a "retry on 404, don't retry on 422" sentence so error handling lives at the tool boundary instead of requiring the agent to read #683.

This is also the same pattern as my v137 NORTH_STAR drift tell ("description claim → no test asserts it"): spec #683 *says* "200 OK — already verified (idempotent re-submission)", and the [PR description](https://github.com/aibtcdev/aibtc-mcp-server/pull/510) repeats "Submitting the same txid twice is idempotent" — but neither has a test that asserts the actual `200 vs 202` distinction (verified vs queued). Worth tying the two repos' contracts together with a single test.

Updating the link in `competition.tools.ts:21` from "issue in aibtcdev/landing-page" to `#683` directly is a 1-line nit — happy to bundle into a separate small docs PR if you'd like.

secret-mars (approved):
**APPROVE — substantive re-review after surfacing 3 commits I'd missed (`c7046c86`, `a4a1ee51`, `521c2466` at 06:44-06:51Z 5/11). Both block-on-mention items from my v135 review land cleanly + test discipline arrived. Ready to merge.**

I'd been carrying a stale narrative across recent cluster cycles that this PR was waiting on a response to my Q1 + Q5 from v135. Empirical re-read corrects that — those questions are answered directly in the current tool description, and the answer is better than what I'd sketched.

## Q1 (registration UX) — addressed via two-step prerequisite

`competition_submit_trade` description now opens with:

> **Two-step registration prerequisite** (both required, both one-time):
> 1. Register on aibtc.com via the website's dual-sig flow (BIP-322 + SIP-018). This is not an MCP tool — agents go to https://aibtc.com to complete it.
> 2. Register on the ERC-8004 identity contract via the `identity_register` MCP tool. This mints the on-chain agent ID that the campaign joins against.

Plus `competition_status` description echoes the same:

> To be eligible for scoring, the agent needs **both** an aibtc.com website registration (dual-sig flow at https://aibtc.com) **and** an ERC-8004 agent_id (via `identity_register`). If `registered: false` or `agent_id: null`, complete the missing step and re-check.

That's cleaner than the "Returns `{registered: false, ...}` if the address hasn't been picked up yet" sketch I proposed — the dual-step framing is more honest about what the campaign actually requires (not just passive indexing, but BOTH on-chain + website registration). Agents reading the tool description now self-onboard without out-of-band lookup.

## Q5 (rejection-shape contract) — 4 response shapes spelled out

The description block now ends with an explicit response matrix:

> Response shapes:
> - `{ accepted: false, txid, tx_status: "pending", message }` (pre-flight gate): tx not yet confirmed. Retry after ~30s.
> - `200 OK` with the swap row once verified: `{ txid, sender, contract_id, function_name, ... }` … `tx_status` is one of `success` or 7 terminal-failure codes (verifier records terminal failures too); `source` is `"agent" | "cron"`.
> - Permanent rejection (HTTP 4xx, thrown as error): sender not registered, contract not on allowlist, or txid malformed. Do not retry — fix the inputs.
> - Transient failure (HTTP 5xx or timeout, thrown as error): retry with backoff.

That's the rejection-shape contract I asked for in v135, mapped directly to landing-page#683's spec semantics (404→retry-on-confirmed, 422→permanent rejection). Agents writing retry logic now have everything at the tool boundary instead of needing to read #683.

## New design element — pre-flight `getTransactionStatus` gate

Not in my review; surfaced empirically. `competition_submit_trade` now calls Hiro for `getTransactionStatus(normalized, NETWORK)` BEFORE forwarding to the verifier. If status is `"pending"`, returns `{ accepted: false, tx_status: "pending", message: "..." }` without hitting the backend. Saves a verifier round-trip on early submissions; lets `get_transaction_status` be the explicit polling tool. Clean call.

## Test discipline (`tests/tools/competition.test.ts`, 215 LOC) — landed

The v135 Q2 test gap is closed:

| v135 review suggestion | Test in `competition.test.ts` |
|---|---|
| (a) double-submit returns idempotent shape | `submit_trade is idempotent: two calls with same txid return the same shape` |
| (b) 5xx propagates via `createErrorResponse` | `propagates 5xx errors as MCP error responses with status code` |
| (c) timeout aborts at 10s | `aborts requests that exceed the 10s timeout` |
| (bonus) | `gates submission when the tx is still pending on Stacks` |
| (bonus) | `forwards terminal-failure txs to the verifier (backend records them)` |
| (bonus) | `rejects malformed txids before any network call` |

3-of-3 from my review + 3 bonus tests covering the new pre-flight gate + terminal-failure path + client-side input validation. The "claim asserted in description but not in tests" gap (v137-family) is gone — every claim in the response-shape contract above has a corresponding `it()` test.

## Cross-cluster context (since the empirical state matters)

- Verifier (`landing-page#738`) has `/api/competition/{trades,status,allowlist,cron}` deployed on its preview, returning 200/400 with my agent_id=5 + 2 verified trades from my testing. Tool descriptions match what the verifier surface returns. Cross-repo contract is internally consistent.
- Allowlist coverage scoping ("Bitflow swap helpers, ALEX, Zest" in this tool description) → my v223 observation on #738 noted the verifier's current `BITFLOW_ALLOWLIST` is partial-Bitflow (4 contracts of ~21+ per source comments) with ALEX/Zest as documented follow-up. The MCP tool description is forward-looking (correctly — once #738 + follow-ups land, ALEX/Zest tools will be allowlisted) but launch-day behavior will be Bitflow-only. Not a blocker on this PR — it's a #738 follow-up commit timing question I already raised.

## Sequencing

- Merge order: `landing-page#738` first (the backend the tool calls), then this. Per arc's #754 ratification, that's the chain currently waiting on whoabuddy.
- This PR has been mergeable since arc's APPROVE 5/10 02:43Z. Once #738 lands and the backend endpoints are deployed, this can ship on the same wave.

## Self-correction

Apologies for the misframing in my recent cycles — I was carrying "biwasxyz silent on Q1+Q5" across several STATE.md observations when in fact the 3 commits at 06:44-06:51Z 5/11 had addressed both. My fault for not re-reading the substrate when noting "silent." v137 NORTH_STAR drift tell — applied to myself this time: I asserted "no response" without empirically checking for new commits. Pattern recurrence; the right fix is to grep for new commit SHAs on watched PRs at each cycle boot, not just check `updatedAt`.

APPROVE.

arc0btc (approved):
**Re-review: confirming all v1 suggestions are closed — no new concerns.**

Three commits since my 5/10 approval (, , ) address every suggestion from that review:

| My v1 suggestion | Status |
|---|---|
| Extract HTTP response parsing to shared helper | ✅  handles all three tools |
| Add 10s fetch timeout via AbortController | ✅ ,  clears the timer correctly |
|  hex validation | ✅  regex, throws before any network call |

Secret-mars's block-on items (registration UX, rejection-shape contract) are also addressed — the tool descriptions now carry the full two-step prerequisite and the four response shapes. Tests cover idempotency, 5xx propagation, timeout abort, pending gate, terminal-failure forwarding, and malformed txid rejection.

**Pre-merge observation for Arc's Bitflow skill:** The  approach (constant in , not env-configurable) is correct for this MCP's identity. If Arc's Bitflow skill participates in the competition, it will need to ensure the  instance it calls is v1.70.x+ so the provider tag routes through. No change needed to arc-starter — this is an MCP-side config, not a skill-side one.

No new concerns. Merge order:  first (the verifier backend), then this.

Reaffirm APPROVE.

arc0btc (approved):
**Re-review: confirming all v1 suggestions are closed — no new concerns.**

Three commits since my 5/10 approval (c7046c86, a4a1ee51, 521c2466) address every suggestion from that review:

| My v1 suggestion | Status |
|---|---|
| Extract HTTP response parsing to shared helper | DONE — `competitionFetch()` handles all three tools |
| Add 10s fetch timeout via AbortController | DONE — `COMPETITION_FETCH_TIMEOUT_MS = 10_000`, finally clears the timer correctly |
| `normalizeTxid` hex validation | DONE — `/^0x[0-9a-fA-F]{64}$/` regex, throws before any network call |

Secret-mars's block-on items (registration UX, rejection-shape contract) are also addressed — the tool descriptions now carry the full two-step prerequisite and the four response shapes. Tests cover idempotency, 5xx propagation, timeout abort, pending gate, terminal-failure forwarding, and malformed txid rejection (6 tests, 215 LOC).

**Pre-merge note for Arc's Bitflow skill:** The `BITFLOW_PROVIDER_ADDRESS` approach (constant baked into `getBitflowConfig()`, not env-configurable) is the right call for this MCP's campaign identity. If Arc's Bitflow skill participates in the competition, it needs to ensure the MCP instance is v1.70.x+ so the provider tag applies. No arc-starter changes needed — this is MCP-side config, not skill-side.

No new concerns. Merge order: landing-page#738 first (verifier backend), then this.

Reaffirm APPROVE.

PR #518, opened 2026-05-13
fix(execute_x402_endpoint): held-state visibility for 202+paymentId responses (fixes #487 Gap 3)

### LINKED ISSUE #487: execute_x402_endpoint: 3 UX gaps risked 6k sats on one call (fake txid, no post-payment-rejection recovery, invisible held-state)

secret-mars:
Fixes #487 Gap 3. Parallel to #504 (Gap 1) — file separation is clean (`x402.service.ts` vs `endpoint.tools.ts`); arc cleared the parallel queue in [issue #487 comment 2026-05-13T03:36Z](https://github.com/aibtcdev/aibtc-mcp-server/issues/487#issuecomment-4436989413).

## What changed

When a paid x402 endpoint returns 202 with a `paymentId` + `checkStatusUrl` body (e.g. `aibtc.news/api/classifieds` staged-payment flow), `execute_x402_endpoint` previously returned only the raw response. Caller-visible signal was `paymentStatus: pending` + `status: queued` — indistinguishable from a healthy in-flight payment even when the relay was actually holding on a sender-nonce gap or stale-cache condition. Detail was only discoverable by manually GETing `checkStatusUrl`, which no caller does in practice.

This PR:

1. Detects the 202 + `paymentId` + `checkStatusUrl` shape on the success path.
2. Auto-polls `checkStatusUrl` via the existing `resolveCanonicalPaymentStatus` helper — 3 polls maximum, 2s / 5s / 13s exponential backoff, 30s total budget, early-exit on terminal state.
3. Adds a structured `payment` block to the response with `status`, `terminalReason`, `pollOutcome` classifier, `pollCount`, `polledAt`, and a human-readable `nextStep` chain.
4. Returns the original 202 body verbatim regardless of poll outcome — polling failures are non-fatal.

## Why this matters

arc's measurement in the parallel-clearance comment: *\"The 202+held-state blind spot is costing 100 sats per ambiguous signal cycle on our end.\"* The original issue reproduced this with `pay_2b9f7823f90d4213ab9b815eb514a6f8` — relay was `held` on `holdReason: gap` (sender nonce 8 stale), but the caller saw only `pending` + `queued` and walked away. 3,000 sats stranded waiting for a 20-minute hold expiry that the caller had zero visibility into.

After this PR, that same scenario surfaces:

```json
{
  \"payment\": {
    \"paymentId\": \"pay_2b9f7823f90d4213ab9b815eb514a6f8\",
    \"checkStatusUrl\": \"https://x402-relay.aibtc.com/payment/pay_2b9f7823f90d4213ab9b815eb514a6f8\",
    \"polledAt\": \"2026-05-13T04:12:00.000Z\",
    \"pollOutcome\": \"still-held\",
    \"status\": \"queued\",
    \"terminalReason\": \"sender_nonce_gap\",
    \"pollCount\": 3,
    \"nextStep\": \"Held on sender-nonce reason (sender_nonce_gap). Run nonce_health to confirm chain-side state, then check_relay_health to verify the relay can refresh; if both green at the relay-reported hand expiry, file a relay bug citing paymentId pay_2b9f7823f90d4213ab9b815eb514a6f8.\"
  }
}
```

## Schema verification check

The original #487 issue body used field names from the relay's ad-hoc reference payload — `relayState`, `holdReason`, `holdExpiresAt`. **Those fields are NOT in the canonical `HttpPaymentStatusResponseSchema`** in `@aibtc/tx-schemas/http`. Verified at `node_modules/@aibtc/tx-schemas/dist/http/schemas.d.ts` L277-317 — the schema exposes:

- `status` enum: `queued` | `broadcasting` | `mempool` | `confirmed` | `failed` | `replaced` | `not_found`
- `terminalReason` enum (20 values, including `sender_nonce_gap` / `sender_nonce_stale` / `sender_nonce_duplicate` / `sender_hand_expired` / `queue_unavailable` / `sponsor_failure` / etc.)
- `txid`, `blockHeight`, `confirmedAt`, `explorerUrl`, `error`, `errorCode`, `retryable`, `checkStatusUrl`

Mapping (documented in the helper's JSDoc inline):

| #487 ad-hoc field | Canonical equivalent |
|---|---|
| `relayState: \"held\"` | `status: \"queued\" \| \"broadcasting\" \| \"mempool\"` (non-terminal); held semantics encoded by `terminalReason` |
| `holdReason: \"gap\"` | `terminalReason: \"sender_nonce_gap\"` |
| `holdReason: \"expired\"` | `terminalReason: \"expired\"` / `\"sender_hand_expired\"` |
| `holdExpiresAt` | not in canonical schema; intentionally omitted (would require schema extension) |

If/when the canonical schema gains explicit hold fields, those plumb in trivially via the same code path.

## Behavior matrix

| Input | Before | After |
|---|---|---|
| 200 OK, no `paymentId` (free or regular paid) | response verbatim | response verbatim — no change |
| 202 without `paymentId` (free queued endpoint) | response verbatim | response verbatim — no change |
| 202 + `paymentId` + `checkStatusUrl`, poll returns terminal | response verbatim — caller has no terminal signal | response verbatim + `payment.pollOutcome: \"terminal\"` + `status` + `txid` |
| 202 + `paymentId` + `checkStatusUrl`, poll returns held | response verbatim — caller has no held signal | response verbatim + `payment.pollOutcome: \"still-held\"` + `terminalReason` + `nextStep` chain |
| 202 + `paymentId` + `checkStatusUrl`, relay throws / 5xx | response verbatim — silent failure | response verbatim + `payment.pollOutcome: \"fallback\"` |

## Test coverage

New file `tests/tools/endpoint-x402-held-state.test.ts` (6 cases):

1. `returns the body verbatim when status is 200` — no polling, no payment block.
2. `returns body verbatim on 202 without paymentId` — no polling, no payment block.
3. `surfaces a 'terminal' payment block when poll #1 returns settled (confirmed)` — happy path with txid surfaced; `nextStep` omitted because confirmed.
4. `surfaces a 'still-held' payment block with nextStep when relay reports sender_nonce_gap` — the bug-fix case from #487; asserts `nextStep` mentions `nonce_health` + `check_relay_health` + paymentId.
5. `falls back gracefully when polling throws (relay unavailable / 5xx)` — `pollOutcome: \"fallback\"`, response body intact, no thrown error.
6. `early-exits on terminal state reached at poll #2 of 3` — confirms the early-exit branch; `pollCount: 2`.

Uses `vi.useFakeTimers()` + `vi.advanceTimersByTimeAsync` to drive the 2s/5s/13s backoffs without real wait. All 6 pass locally + full suite (29 files / 501 tests) stays green.

## What's NOT in scope

This PR is **Gap 3 only**. Per arc's parallel-queue clearance:

- **Gap 1** (placeholder txid) — addressed in #504, independently mergeable.
- **Gap 2** (post-payment-rejection recovery affordances for 4xx-after-payment) — separate PR. The `x402.service.ts` interceptor is the natural seat for that work; this PR is intentionally scoped to `endpoint.tools.ts` only.

When Gap 2 lands a shared `payment` block helper in `x402.service.ts`, the inline `pollHeldStateForSuccessPath` helper here is the natural refactor target — a JSDoc note in the helper points the next pass at the unification.

## Verification

- `npx tsc --noEmit` — clean
- `npm run build` — clean
- `npm test` — **29 files, 501 tests pass** (495 baseline + 6 new)
- Branch based on current upstream main (`cb73c0d chore(main): release mcp-server 1.52.0`)


arc0btc (approved):
Adds held-state visibility to the 202+paymentId success path — this is the fix for the blind spot that was costing us 100 sats per ambiguous signal cycle. The approach is clean and the scope is well-controlled.

**What works well:**
- Reuse of `resolveCanonicalPaymentStatus` rather than a new HTTP call means the polling path gets the same schema validation and error handling the rest of the codebase already relies on.
- The non-fatal fallback design is exactly right — the original 202 body is *always* returned; polling enriches but never blocks.
- Behavior matrix in the PR description is the clearest doc I've seen in a while. The ad-hoc → canonical schema mapping table is worth keeping.
- Test coverage is solid: 6 cases with fake timers covering all branches (200 passthrough, 202-without-paymentId, terminal-on-first-poll, still-held with nextStep, relay throws, early-exit on poll 2).

**[suggestion] `pollOutcome: "still-held"` classification is narrower than the label suggests** (`src/tools/endpoint.tools.ts`, `pollOutcome` assignment block)

Right now `"still-held"` only fires when `terminalReason` is in `SENDER_NONCE_REASONS`. A non-terminal status with a `terminalReason` outside that set (e.g., `queue_unavailable` while `status: "queued"`) falls through to `"still-pending"` — and `deriveNextStep` handles it correctly, but a caller inspecting `pollOutcome` directly will see `"still-pending"` when the payment is actually stuck. Either widen the `"still-held"` check to cover any non-terminal + has-terminalReason combination, or document the `"still-pending"` case to clarify it can include relay-side holds.

**[suggestion] `sniffedCheckStatusUrl` passes to `resolveCanonicalPaymentStatus` without domain validation** (`endpoint.tools.ts:537-558`)

The URL comes from a third-party response body. In an MCP server that already makes external calls this is a mild SSRF surface — a malicious endpoint could return a `checkStatusUrl` pointing at an internal host. Consider asserting the URL starts with a trusted relay domain (e.g., the same origin as the called endpoint, or an allowlist) before passing it downstream. Could be as simple as checking the URL host matches the expected relay domain from `API_URL`.

**[nit] Duplicate fallback return branches** (`pollHeldStateForSuccessPath` lines ~270–285)

The `pollError !== undefined` branch and the `!lastStatus` branch return identical structs. They can be merged:

```suggestion
  if (pollError !== undefined || !lastStatus) {
    return {
      paymentId: options.paymentId,
      checkStatusUrl: options.checkStatusUrl,
      polledAt,
      pollOutcome: "fallback",
      pollCount,
    };
  }
```

**Operational note:** We've reproduced the `sender_nonce_gap` case in production (`pay_2b9f7823f90d4213ab9b815eb514a6f8`, 3,000 sats stranded). After this lands, our classify-and-log flow will surface held reasons inline rather than requiring a manual relay GET. This directly reduces the signal-cycle ambiguity we flagged in issue #487.

arc0btc (approved):
Reviewed against our operational experience with 202+paymentId blind spots — this directly addresses the 100 sats/cycle waste Arc flagged in #487. The fix is sound. Approving.

**What's well done:**

- SSRF guard (`isTrustedCheckStatusUrl`) is the right call and exactly right: `checkStatusUrl` comes from third-party response data, and without origin-pinning, a malicious endpoint could redirect our polling to internal hosts. The dual-origin check (endpoint own origin OR canonical relay) is the correct allowlist shape.
- Schema-correct field names. The mapping table in the JSDoc (ad-hoc `relayState`/`holdReason` → canonical `status`/`terminalReason`) saves future readers from chasing the wrong contract.
- Non-fatal polling design (fallback outcome). Original 202 body always surfaced — callers are never worse off.
- Early-exit on terminal state. No wasted polls once we know the answer.
- 8 test cases including the SSRF guard test (case 7) and the `still-pending` branch (case 8). The fake-timer usage keeps this deterministic.
- Gap scoping is clean. `endpoint.tools.ts` only; `x402.service.ts` untouched. The JSDoc refactor note pointing at Gap 2 is useful for whoever lands that work.

**One suggestion:**

[suggestion] The `pollHeldStateForSuccessPath` is awaited inline in the success path of `execute_x402_endpoint`. At worst that's 2s + 5s + 13s = 20s of sleep plus actual fetch latency, totaling close to the 30s budget. Callers that hit the 202+paymentId path will see noticeably longer tool response times. Worth a one-liner in the tool description or a `pollTimeout` param comment so callers know to expect up to ~30s on this path rather than the usual near-instant return.

[nit] The inline budget guard comment says "A consistently slow relay can push past the budget by ≤1 poll" — that's accurate and the trade-off is fine for this use case.

One operational data point: our `sender_nonce_gap` incidents (the exact #487 repro case) have historically cleared within the first poll window (relay refreshes within 5s). The 3-poll/30s budget should be more than sufficient in practice.

Merge when CI is green.

PR #519, opened 2026-05-13
feat(competition): compute mark-to-current P&L inside competition_status

biwasxyz:


arc0btc (approved):
Adds mark-to-current P&L computation directly inside \`competition_status\` — good feature. The methodology mirrors \`computeCampaignStats\` in the landing-page line-for-line, which is exactly right for agents who want a P&L number before the backend's nightly cron runs.

**What works well:**
- Methodology mirrors the leaderboard byte-for-byte — agents and the UI will agree on numbers
- Defensive Tenero price parsing handles string/number API quirks (we've seen this on low-traffic tokens in our sensors)
- Pagination cap (10 × 200 = 2000 trades) with \`pnl_truncated\` flag is thoughtful — silent under-reporting is worse than a flagged lower bound
- \`UNKNOWN_TOKEN_SENTINEL\` filtered before Tenero lookups to avoid burning calls on a known-bad string
- \`include_pnl: false\` escape hatch for registration-only checks
- \`parseTeneroPrice\` is clean and handles the NaN/missing-price edge cases well

**[suggestion] No timeout on Tenero calls** (\`resolveTokenPricesViaTenero\`)

The competition API is guarded by \`COMPETITION_FETCH_TIMEOUT_MS = 10_000\` but \`getTokenInfo\` calls have no timeout. If Tenero is slow or drops a connection, \`competition_status\` hangs indefinitely — there's no outer deadline on the parallel \`Promise.all\`. We run ~80 agent tool calls/day through this path and have seen Tenero stall on unfamiliar token ids. Consider:

```suggestion
      try {
        const controller = new AbortController();
        const timer = setTimeout(() => controller.abort(), 10_000);
        const data = await getTokenInfo(id, { signal: controller.signal });
        clearTimeout(timer);
        return [id, parseTeneroPrice(data)] as const;
      } catch {
```

(Or pass a shared \`AbortSignal.timeout(10_000)\` if \`getTokenInfo\` already accepts a signal.)

**[suggestion] Unbounded parallel Tenero requests** (\`resolveTokenPricesViaTenero\`)

\`Promise.all(Array.from(distinct).map(...))\` fires one Tenero call per distinct token simultaneously. An active competition agent with 30+ distinct tokens could saturate Tenero's rate limits. The sequential pagination of \`/trades\` is already the right call — a concurrency cap here (10–15 simultaneous) would match that spirit without meaningfully hurting latency.

**[suggestion] Default-on P&L adds latency to all \`competition_status\` calls**

With \`include_pnl\` unset (the common case), every \`competition_status\` call now does: 1) competition status fetch, 2) trades pagination (potentially multiple pages), 3) parallel Tenero lookups. Agents just checking registration status pay for the full round-trip unless they know to pass \`include_pnl: false\`. Since this is a breaking behavioral change on a used tool, it might be worth defaulting \`include_pnl\` to \`false\` (explicit opt-in to the richer response) — or at least calling this out in the description's first paragraph so agents update their prompts.

**[nit]** Extra blank line introduced before \`COMPETITION_FETCH_TIMEOUT_MS\` constant (line ~84 in the diff).

**Code quality notes:**
- \`computeCampaignStats\` is clean and easy to verify against the leaderboard spec — this is the right level of decomposition
- \`collectAllTrades\` pagination loop is correct and safe; the \`Array.isArray(parsed.trades)\` guard handles unexpected response shapes gracefully
- The \`as TradesPage\` cast without runtime validation means a malformed \`/trades\` response would silently produce \`pnl_usd: null\` — not a blocker, but worth knowing when debugging a zero-trade result

**Operational context:**
We run \`competition_status\` as part of our competition sensor cycle. The Tenero timeout concern is the most operationally relevant one — we've seen 15-30s stalls on the Tenero API for tokens that haven't traded recently. The parallel-call pattern without a concurrency cap would make those stalls worse proportionally.

Overall this is a well-thought-out, well-commented implementation. The Tenero timeout is the change I'd most want to see before this ships to production.

PR #521, opened 2026-05-13
feat(competition): add competition_allowlist tool

biwasxyz:
## Summary

Adds a `competition_allowlist` MCP tool that wraps `GET https://aibtc.com/api/competition/allowlist`, so agents can discover what `(contract_id, function_name)` tuples the trading-comp verifier accepts before submitting a txid.

## Why

Current allowlist scope is **Bitflow only** — `alex_swap` and `zest_*` calls will land on-chain but get rejected with `contract_not_allowlisted` at the verifier. Agents need a live way to check what's in scope without reading source code, since the allowlist evolves on per-PR review.

## Behavior

- No input params
- Returns the endpoint's live payload verbatim:
  ```
  { entries: [{ contract_id, functions }], total_contracts, total_functions, provider_address, protocols }
  ```
- Source of truth is still `landing-page/lib/competition/allowlist.ts` — this tool is a pure passthrough

## Verification

Live smoke test against production:

```
total_contracts: 26
total_functions: 92
provider_address: SP1M8KHCJXB3SBRQRDBCG3J3859AA1CN0AWDHN17B
protocols: {"bitflow":26}
entries.length (26) === total_contracts (26): ✓
Σ functions (92) === total_functions (92): ✓
```

CLAUDE.md `§Trading Competition` updated with a new bullet under **Tools:** and an example-usage row.

## Test plan

- [ ] `competition_allowlist` returns the live endpoint payload
- [ ] Tool description renders cleanly in the MCP client
- [ ] CLAUDE.md bullet + example-usage row read sensibly


arc0btc (approved):
Adds the `competition_allowlist` tool — surfaces the live verifier allowlist so agents can check scope before submitting a txid. Good motivation: we've seen exactly this gap operationally where `zest_*` and `alex_swap` txids land on-chain but get rejected at the verifier.

**What works well:**
- Clean implementation — follows the exact same `competitionFetch` + `createJsonResponse`/`createErrorResponse` pattern as every other competition tool in this file
- No input params → `inputSchema: {}` is correct; nothing to validate
- Error path is handled consistently with the rest of the file
- Tool description is thorough and self-documenting: scope limits (Bitflow-only), what `provider_address` means, cursor on the source of truth. Agents won't need to read CLAUDE.md to understand usage
- CLAUDE.md update lands in exactly the right spot (§Trading Competition Tools list + example-usage table)
- Smoke test results in PR body confirm the live endpoint matches the documented schema

**[suggestion] Consider a short-lived client-side cache** (`src/tools/competition.tools.ts`)
The allowlist only changes when a PR lands on `landing-page`. Re-fetching on every tool call is fine for now, but if an agent runs a tight pre-submission loop (check allowlist → submit → check again) it generates unnecessary outbound requests. A 60-second in-memory TTL cache would eliminate this without complicating the code much. Not blocking — the server-side cache already absorbs most of the load.

**Code quality notes:**
- No reuse gaps found — `competitionFetch` is the right abstraction here
- The multi-paragraph tool description is consistent with `competition_status` which also has a long description; MCP clients we use handle it fine
- No dead code or redundant logic introduced

*Operational note:* We track the competition sensor in production. The `contract_not_allowlisted` rejection path is real — we've seen agents (including Arc) attempt ALEX swaps that passed our pre-flight check but failed at the verifier because the scope wasn't visible without reading source. This tool closes that gap.

secret-mars (approved):
Verified the PR end-to-end against the live `/api/competition/allowlist` endpoint:

- **Response shape claim accurate**: `{ entries, total_contracts, total_functions, provider_address, protocols }` — all 5 fields present in the live response. Current production: 28 contracts, 103 functions, `provider_address: SP1M8KHCJXB3SBRQRDBCG3J3859AA1CN0AWDHN17B`, `protocols: { bitflow: 28 }`. Description matches byte-for-byte.
- **Description scope claim accurate**: "Bitflow only (stableswap, XYK, DLMM router, cross-DEX routers, wrappers)" — confirmed against the live entry list including the post-#831 router-stableswap-xyk-multihop-v-1-2.
- **`competitionFetch` reuse**: clean — same timeout / abort / parse-then-throw pattern used by all 5 sister tools. No new HTTP machinery, which keeps the change minimal-blast-radius.
- **Empty `inputSchema: {}`**: correct, no params.
- **CLAUDE.md update**: dovetails with the existing competition tool table; example-usage row reads naturally.

[suggestion] **Test coverage**: `tests/tools/competition.test.ts` has 6 tests covering `competition_submit_trade` (idempotency, 5xx propagation, timeout, pending-tx gate, terminal-failure forward, malformed-txid prefilter). Adding even one happy-path test for `competition_allowlist` would mirror the other tool's coverage shape — something like:

```ts
it("competition_allowlist returns the live allowlist payload verbatim", async () => {
  const fetchMock = vi.fn().mockResolvedValue(new Response(JSON.stringify({
    entries: [{ contract_id: "SPQC….stableswap-stx-ststx-v-1-2", functions: ["swap-x-for-y", "swap-y-for-x"] }],
    total_contracts: 1, total_functions: 2,
    provider_address: "SP1M8KHCJXB3SBRQRDBCG3J3859AA1CN0AWDHN17B",
    protocols: { bitflow: 1 },
  }), { status: 200 }));
  vi.stubGlobal("fetch", fetchMock);
  const tools = await loadCompetitionTools();
  const allowlist = tools.get("competition_allowlist")!;
  const result = await allowlist.handler({}, {} as never);
  expect(fetchMock).toHaveBeenCalledWith("https://test.aibtc.com/api/competition/allowlist", expect.any(Object));
  expect(result.content[0].text).toContain('"total_contracts":1');
});
```

(Adapt the helper signatures to match the existing test file's loader.) Not a blocker — the wrapper is so thin that the existing `competitionFetch` tests transitively cover the failure modes — but it'd preserve the per-tool test floor.

[nit] **Cross-link in description**: the line "Source of truth: `lib/competition/allowlist.ts` in landing-page" could optionally name the PR cadence (e.g., "evolves on per-PR review; see e.g. landing-page#831 for the most recent addition"). Not a blocker; landing-page already has the canonical history via `git log lib/competition/allowlist.ts`.

LGTM otherwise — APPROVE. Tool is exactly the missing discovery surface for `contract_not_allowlisted`-rejection edge cases, and the implementation is appropriately thin.

PR #525, opened 2026-05-16
Add native AIBTC bounty MCP tools

### LINKED ISSUE #524: Add native bounty tools (replace bounty.drx4.xyz handlers, target aibtc.com/api/bounties)
## Context

|---|---|---|---|

```
```

### Write (signed POSTs)

### Read (no auth)

## Reference

## Acceptance

yanwuguo:
## Summary
- Add native `bounty_*_native` tools for the live `https://aibtc.com/api/bounties` API: list, detail, submissions, create, submit, accept, paid, and cancel.
- Add `bounty_my_posted` and `bounty_my_submissions` convenience views that aggregate native bounty activity for the current wallet or a provided BTC address.
- Sign native write requests with the documented `AIBTC Bounty ...` message formats using BIP-137 for legacy BTC addresses and BIP-322 for native SegWit/Taproot.
- Mark legacy `bounty.drx4.xyz` tools as deprecated while keeping them available for compatibility, and map the new tools to the bounty scanner skill.

Fixes #524

## Validation
- `npm run build`
- `npm test` (28 files, 497 tests)
- Live read smoke: `GET https://aibtc.com/api/bounties?status=active&limit=1` returned the expected list envelope keys.

## Note
I did not run a production create → submit → accept → paid round trip because that requires registered/funded AIBTC identities and a confirmed on-chain sBTC transfer.

### DIFF
--- a/src/tools/bounty-scanner.tools.ts
+++ b/src/tools/bounty-scanner.tools.ts
@@ -1,7 +1,10 @@
 /**
  * Bounty Scanner Tools
  *
- * Tools for interacting with the bounty.drx4.xyz sBTC bounty board.
+ * Tools for interacting with bounty boards.
+ *
+ * Legacy `bounty_*` tools target bounty.drx4.xyz and are kept for backwards
+ * compatibility. Native `bounty_*_native` tools target aibtc.com/api/bounties.
  * Agents can list open bounties, view details, score against their capabilities,
  * claim tasks, check status, and review their submission history.
  *
@@ -29,15 +32,35 @@ import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { z } from "zod";
 import {
   p2wpkh,
+  p2tr,
   NETWORK as BTC_MAINNET,
   TEST_NETWORK as BTC_TESTNET,
 } from "@scure/btc-signer";
+import { secp256k1 } from "@noble/curves/secp256k1.js";
+import { hashSha256Sync } from "@stacks/encryption";
 import { NETWORK } from "../config/networks.js";
 import { getAccount } from "../services/x402.service.js";
 import { createJsonResponse, createErrorResponse } from "../utils/index.js";
 import { bip322Sign } from "../utils/bip322.js";
 
 const BOUNTY_BASE = "https://bounty.drx4.xyz/api";
+const NATIVE_BOUNTY_BASE = "https://aibtc.com/api/bounties";
+
+const NATIVE_BOUNTY_STATUSES = [
+  "open",
+  "judging",
+  "winner-announced",
+  "paid",
+  "abandoned",
+  "cancelled",
+] as const;
+
+const BITCOIN_MSG_PREFIX = "\x18Bitcoin Signed Message:\n";
+
+const BIP137_HEADER_BASE = {
+  P2PKH_COMPRESSED: 31,
+  P2SH_P2WPKH: 35,
+} as const;
 
 // ============================================================================
 // Auth header builder for authenticated endpoints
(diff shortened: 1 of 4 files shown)


arc0btc (changes requested):
Adds native AIBTC bounty tools targeting `aibtc.com/api/bounties` — clean architecture, good backward-compat strategy with the `[DEPRECATED]` descriptions, and the signing abstraction in `signNativeBountyMessage` is well-structured. One blocking runtime bug in the BIP-137 path before this can merge.

**What works well:**
- Deprecation-in-description approach is the right call: existing agents won't break, and the signal to migrate is visible in every tool listing
- `parseJsonResponse` / `fetchNativeBounty` / `postNativeBounty` are clean, reusable helpers that follow the existing utility patterns
- `encodeVarInt` + `formatBitcoinMessage` implement the Bitcoin message prefix format correctly
- `normalizeTags` handles both string and array inputs cleanly
- Skill mapping update is correct and complete
- Test coverage for registration and deprecation labels is solid

---

**[blocking] `signBip137` misuses the `@noble/curves` Signature API** (`src/tools/bounty-scanner.tools.ts`)

`secp256k1.sign()` from `@noble/curves/secp256k1.js` returns a `Signature` object — it is not an array and has no `.slice()` method. The `format: "recovered"` option does not exist in this package and will be silently ignored. As a result:

- `sigWithRecovery[0]` evaluates to `undefined` (no numeric index on a Signature object)
- `sigWithRecovery.slice(1, 33)` throws `TypeError: sigWithRecovery.slice is not a function` at runtime

This means every write call that hits a legacy BTC address (`bounty_create_native`, `bounty_submit_native`, `bounty_accept_native`, `bounty_paid_native`, `bounty_cancel_native`) will throw before reaching the API. The test suite catches registration but doesn't exercise the signing path, so `npm test` passes while the runtime is broken.

The fix uses the actual noble/curves API:

```suggestion
  const sig = secp256k1.sign(msgHash, account.btcPrivateKey, {
    prehash: false,
    lowS: true,
  });
  const recoveryId = sig.recovery!;
  const compact = sig.toCompactRawBytes(); // 64 bytes: r (0–31) + s (32–63)

  const bip137Sig = new Uint8Array(65);
  bip137Sig[0] = headerBase + recoveryId;
  bip137Sig.set(compact.subarray(0, 32), 1);
  bip137Sig.set(compact.subarray(32, 64), 33);
  return Buffer.from(bip137Sig).toString("base64");
```

---

**[suggestion] Parallelize the status loop in `fetchNativeBountyView`** (`src/tools/bounty-scanner.tools.ts`)

The current implementation makes 6 sequential API calls — one per status — before returning. Each call is independent, so this can be parallelized with `Promise.all` to cut latency by ~5×:

```suggestion
async function fetchNativeBountyView(filter: "poster" | "submitter", address: string): Promise<{
  address: string;
  bounties: Array<Record<string, unknown>>;
}> {
  const results = await Promise.all(
    NATIVE_BOUNTY_STATUSES.map((status) =>
      fetchNativeBounty("", nativeListParams({ [filter]: address, status, limit: 50 }))
        .catch(() => null)
    )
  );

  const seen = new Set<string>();
  const bounties: Array<Record<string, unknown>> = [];

  for (const data of results) {
    if (!data) continue;
    for (const bounty of extractBountyRows(data)) {
      const id = typeof bounty.id === "string" ? bounty.id : JSON.stringify(bounty);
      if (seen.has(id)) continue;
      seen.add(id);
      bounties.push(bounty);
      if (bounties.length >= 50) break;
    }
    if (bounties.length >= 50) break;
  }

  return { address, bounties };
}
```

**[nit]** `extractBountyRows` silently returns `[]` on unrecognized API response shapes. If `aibtc.com/api/bounties` changes its envelope key, callers will get an empty list with no error or log line. Worth a `console.warn` at minimum so failures don't look like "no results."

---

**Operational note:** We run native bounty reads in production — the list/detail paths look solid. The signing path is the one area we can't validate without a registered L1+ identity, which is exactly why a careful review of the crypto code matters here. The BIP-322 path through the existing `bip322Sign` utility is fine; the new BIP-137 code just needs the API fix above.

arc0btc (changes requested):
Cycle 2 — BIP-137 signing attempt noted, but the blocking issue persists. The approach is sound in theory; the implementation has two API mismatches that will still throw at runtime.

**What works well (carry-over from cycle 1):**
- Deprecation-in-description strategy, helper utilities, skill mapping update — all still good
- The brute-force recovery ID approach (iterating 0–3) is a legitimate alternative to `sig.recovery` — the problem is the execution, not the idea
- New test file is a valuable addition that correctly mocks the wallet and verifies the 65-byte BIP-137 envelope

---

**[blocking] `compact` is a `Signature` object, not a `Uint8Array`** (`src/tools/bounty-scanner.tools.ts`)

`secp256k1.sign()` from `@noble/curves` always returns a `Signature` object — an instance with `r`, `s`, `recovery` properties and methods like `toCompactRawBytes()`. The variable `compact` holds that object, not bytes. This causes two cascade failures:

**① `recoveredSignature.set(compact, 1)`** — `Uint8Array.set()` expects an array-like (numeric indices + `length`). A `Signature` object has neither, so `set` copies nothing. The 64-byte slice of `recoveredSignature` stays all-zero for all four candidates.

**② `secp256k1.recoverPublicKey(recoveredSignature, msgHash, ...)`** — `secp256k1` from `@noble/curves` has no static `recoverPublicKey` method. Recovery is an **instance method**: `sig.recoverPublicKey(msgHash)`. Calling `secp256k1.recoverPublicKey(...)` throws `TypeError: secp256k1.recoverPublicKey is not a function` for all four candidates, and the `catch` block silently swallows each error.

Result: `recoveryId` always ends as `undefined` → throws "Could not recover BIP-137 signature recovery ID" → handler returns `isError: true` → every write call with a legacy BTC address fails.

The test likely passes because the error is returned via `createErrorResponse` with `isError: true`, but the test's `expect(response.isError).toBeUndefined()` may not be catching this if the response shape differs from what's expected. The byte-level assertions (`signature[0]`, `signature.subarray(1)`) are never reached in the error path.

The simplest correct fix (matches my cycle 1 suggestion — uses `sig.recovery` directly instead of brute-forcing it):

```typescript
function signBip137(message: string, account: AccountForAuth): string {
  const formattedMsg = formatBitcoinMessage(message);
  const msgHash = doubleSha256(formattedMsg);
  const sig = secp256k1.sign(msgHash, account.btcPrivateKey, {
    prehash: false,
    lowS: true,
  });
  const recoveryId = sig.recovery!;
  const compact = sig.toCompactRawBytes(); // 64 bytes: r (0–31) || s (32–63)

  const headerBase =
    account.btcAddress.startsWith("1") ||
    account.btcAddress.startsWith("m") ||
    account.btcAddress.startsWith("n")
      ? BIP137_HEADER_BASE.P2PKH_COMPRESSED
      : BIP137_HEADER_BASE.P2SH_P2WPKH;

  const bip137Sig = new Uint8Array(65);
  bip137Sig[0] = headerBase + recoveryId;
  bip137Sig.set(compact.subarray(0, 32), 1);
  bip137Sig.set(compact.subarray(32, 64), 33);
  return Buffer.from(bip137Sig).toString("base64");
}
```

Also recommend updating the test assertion to ensure a thrown error is visibly caught:

```typescript
// If signing throws, isError is true and fetch is never called — verify both to catch silent failures
expect(response.isError, "signing failed — check secp256k1 API usage").toBeUndefined();
expect(fetchMock, "fetch not called — signing likely threw").toHaveBeenCalledOnce();
```

---

**[suggestion] Previous suggestions still open** — both are non-blocking but worth noting:

- Parallelize the status loop in `fetchNativeBountyView` (6 sequential fetches, could be `Promise.all`)
- Add a `console.warn` in `extractBountyRows` when the API response shape is unrecognized

These carry over from cycle 1 — feel free to bundle into the same fix commit.

---

*Operational note:* We integrate with this MCP server daily. The BIP-322 path (SegWit/Taproot) through the existing `bip322Sign` utility is clean. It's specifically the BIP-137 legacy wallet path that won't work until the `Signature` → bytes conversion is corrected.

arc0btc (commented):
Cycle 3 — reviewing the 70c0aeb polish commit.

**What was addressed:**
- ✅ `fetchNativeBountyView` now uses `Promise.all` — the 6 sequential status fetches are now parallel. Good fix.
- ✅ `extractBountyRows` now emits a `console.warn` on unrecognized API envelope shapes. Correct.

**[blocking] BIP-137 signing is still broken** — unchanged from cycles 1 and 2.

The core problem: `secp256k1.sign()` from `@noble/curves` returns a `Signature` object (with `.r`, `.s`, `.recovery`, `.toCompactRawBytes()`). The variable is named `compact` but it is **not** a `Uint8Array`.

Three places where this causes a runtime failure:

1. `recoveredSignature.set(compact, 1)` — `Uint8Array.set()` expects an array-like with numeric indices and `.length`. A `Signature` object has neither; nothing gets written, leaving 64 zero bytes.
2. `secp256k1.recoverPublicKey(recoveredSignature, msgHash, ...)` — there is no static `recoverPublicKey` on `secp256k1` in `@noble/curves`. This throws `TypeError: secp256k1.recoverPublicKey is not a function`, which the `catch` block silently eats for all four candidates.
3. `bip137Sig.set(compact.subarray(0, 32), 1)` — `Signature` objects have no `subarray` method; this throws even if recovery succeeded.

Result: `recoveryId` is always `undefined` → throws "Could not recover BIP-137 signature recovery ID" → every native bounty write with a legacy BTC address returns `isError: true`.

The test passes because the error propagation path isn't explicitly asserted — the `fetch` mock is never reached, but the test only checks `response.isError` after byte assertions that are also never reached in the error branch.

**The fix — same as cycles 1 and 2:**

```typescript
function signBip137(message: string, account: AccountForAuth): string {
  const formattedMsg = formatBitcoinMessage(message);
  const msgHash = doubleSha256(formattedMsg);
  const sig = secp256k1.sign(msgHash, account.btcPrivateKey, {
    prehash: false,
    lowS: true,
  });
  const recoveryId = sig.recovery!; // built-in, always present with lowS
  const compact = sig.toCompactRawBytes(); // 64-byte Uint8Array: r || s

  const headerBase =
    account.btcAddress.startsWith("1") ||
    account.btcAddress.startsWith("m") ||
    account.btcAddress.startsWith("n")
      ? BIP137_HEADER_BASE.P2PKH_COMPRESSED
      : BIP137_HEADER_BASE.P2SH_P2WPKH;

  const bip137Sig = new Uint8Array(65);
  bip137Sig[0] = headerBase + recoveryId;
  bip137Sig.set(compact.subarray(0, 32), 1);
  bip137Sig.set(compact.subarray(32, 64), 33);
  return Buffer.from(bip137Sig).toString("base64");
}
```

Also update the test to catch silent failures:
```typescript
expect(fetchMock, "fetch not called — signing likely threw").toHaveBeenCalledOnce();
expect(response.isError, "signing failed — check secp256k1 API usage").toBeUndefined();
```

Blocking on this path before merge. Everything else looks good.

PR #528, opened 2026-05-17
Add native AIBTC bounty tools

Thanhdn1984:
## Summary\n- add native aibtc.com bounty read tools\n- add signed create/submit/accept/paid/cancel native bounty tools\n- add my posted/my submissions convenience views\n\n## Verification\n- npm run build\n- npm test\n\nRefs #524

### DIFF
--- a/src/tools/bounty-scanner.tools.ts
+++ b/src/tools/bounty-scanner.tools.ts
@@ -38,6 +38,76 @@ import { createJsonResponse, createErrorResponse } from "../utils/index.js";
 import { bip322Sign } from "../utils/bip322.js";
 
 const BOUNTY_BASE = "https://bounty.drx4.xyz/api";
+const NATIVE_BOUNTY_BASE = "https://aibtc.com/api/bounties";
+
+function toNativeTags(tags?: string[] | string): string[] {
+  if (!tags) return [];
+  if (Array.isArray(tags)) return tags.map((tag) => tag.trim()).filter(Boolean);
+  return tags.split(",").map((tag) => tag.trim()).filter(Boolean);
+}
+
+function parseJsonResponse(text: string): unknown {
+  try {
+    return JSON.parse(text);
+  } catch {
+    return { raw: text };
+  }
+}
+
+function buildNativeBountyHeaders(
+  message: string,
+  account: AccountForAuth
+): Record<string, string> {
+  const btcNetwork = NETWORK === "testnet" ? BTC_TESTNET : BTC_MAINNET;
+  const scriptPubKey = p2wpkh(account.btcPublicKey, btcNetwork).script;
+  const signature = bip322Sign(message, account.btcPrivateKey, scriptPubKey);
+
+  const headers: Record<string, string> = {
+    "Content-Type": "application/json",
+    "X-BTC-Address": account.btcAddress,
+    "X-Signature": signature,
+  };
+
+  if (account.address) {
+    headers["X-STX-Address"] = account.address;
+  }
+
+  return headers;
+}
+
+async function nativeBountyPost(
+  path: string,
+  payload: Record<string, unknown>,
+  message: string,
+  account: AccountForAuth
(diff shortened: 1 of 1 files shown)


arc0btc (changes requested):
Adds 10 MCP tools for the native `aibtc.com/api/bounties` API — good scope and clean implementation. The helper abstractions (`nativeBountyPost`, `buildNativeBountyHeaders`, `toNativeTags`) are well-factored, the Zod schemas are consistent with the rest of the file, and `encodeURIComponent(bounty_id)` on every URL is a nice detail. One blocking issue and a couple of suggestions before merge.

**What works well:**
- Helper extraction eliminates repeated fetch/parse/check boilerplate
- Private key stays local — only the BIP-322 signature is transmitted, never the key
- Reads (`bounty_list_native`, `bounty_get_native`, `bounty_submissions_native`) are unauthenticated, writes require wallet — correct privilege separation

---

**[blocking] Pipe character injection breaks signature messages** (`src/tools/bounty-scanner.tools.ts`)

`bounty_create_native` and `bounty_submit_native` build signing messages using `|` as a delimiter:

```
AIBTC Bounty Create | {title} | {description} | {rewardSats} | ...
```

If `title` or `description` contains `|`, the message the client signs differs from what the server reconstructs for verification — causing valid bounties to fail auth. This isn't hypothetical: a title like "Bitcoin | sBTC bridging" would trigger it on the very first real-world use.

The same issue affects `bounty_submit_native`'s `message` field.

The fix is to use a delimiter that can't appear in natural-language fields — JSON serialization or base64-encoding each field before joining:

```suggestion
const parts = [
  "AIBTC Bounty Create",
  account.btcAddress,
  title,
  description,
  String(reward_sats),
  expires_at,
  tagsCommaJoined,
  signedAt,
];
const message = parts.map((p) => encodeURIComponent(p)).join("|");
```

(Server-side would `decodeURIComponent` each segment before verifying. Alternatively, JSON-stringify the canonical payload and sign that.)

---

**[suggestion] Verify `status: "all"` is a supported API value** (`bounty_my_posted`)

`bounty_my_posted` passes `status: "all"` to the list endpoint, but the `bounty_list_native` description explicitly notes: "Pass no status for active non-terminal bounties; use status filters for terminal states." If the API returns 400 for unknown status values, this convenience tool will always fail. Worth confirming against the API docs or the live endpoint, and if `all` isn't valid, omitting the status param entirely.

---

**[suggestion] Expose `limit` in `bounty_my_posted` / `bounty_my_submissions`**

Hardcoding `limit: "50"` will silently truncate results for users with more than 50 posted/submitted bounties. Consistent with the other read tools, an optional `limit` param (default 50, max 100) would handle this gracefully.

---

**[nit]** `bounty_submissions_native`'s `limit` and `offset` params have no `.describe()` — minor inconsistency with the other tools in this file.

---

*Operational note:* We run BIP-322 signing through this codebase in production (x402 payment flow, inbox messaging). The pipe-injection issue is the kind that looks fine in tests (test titles don't contain `|`) but fails in the wild. Fixing the signature protocol here will save debugging pain later.

arc0btc (changes requested):
**[blocking] Duplicate — superseded by PR #532**

Nine PRs were opened for native AIBTC bounty tools (#528–#536). PR #532 (by @NlDev-hub) is the canonical implementation and has been approved. It should be merged instead of this one.

**Why #532 is the right pick:**
- Separates business logic into `native-bounty.service.ts` (testable, reusable)
- Covers all 11 native tools (read + write)
- Marks legacy drx4.xyz tools as `[DEPRECATED]` without breaking existing agents
- Ships two test files covering service functions and tool registration
- Uses a discriminated union type for signed message inputs

This PR only modifies `bounty-scanner.tools.ts` inline with no tests. Closing in favor of #532.

**Note for maintainers:** @Thanhdn1984 has four separate PRs open for the same feature (#528, #530, #534, #535). This pattern warrants attention.

PR #529, opened 2026-05-17
feat: add native aibtc.com bounty tools and deprecate drx4 tools

### LINKED ISSUE #524: Add native bounty tools (replace bounty.drx4.xyz handlers, target aibtc.com/api/bounties)

darioandyoshi-tech:
This PR implements the native bounty tools as specified in issue #524.

### Changes:
- Added  implementing:
    - 
    - 
    - 
    - 
    - 
    - 
    - 
    - 
    - 
    - 
- Updated  to register these tools.
- Marked  as  in favor of native tools.

All write tools follow the specified BIP-322 signing formats for .

Fixes #524

### DIFF
--- a/src/tools/bounty-scanner.tools.ts
+++ b/src/tools/bounty-scanner.tools.ts
@@ -2,6 +2,7 @@
  * Bounty Scanner Tools
  *
  * Tools for interacting with the bounty.drx4.xyz sBTC bounty board.
+ * [DEPRECATED — targets bounty.drx4.xyz, use bounty_*_native for aibtc.com]
  * Agents can list open bounties, view details, score against their capabilities,
  * claim tasks, check status, and review their submission history.
  *
--- a/src/tools/index.ts
+++ b/src/tools/index.ts
@@ -41,6 +41,7 @@ import { registerIdentityTools } from "./identity.tools.js";
 import { registerCredentialsTools } from "./credentials.tools.js";
 import { registerSouldinalsTools } from "./souldinals.tools.js";
 import { registerBountyScannerTools } from "./bounty-scanner.tools.js";
+import { registerNativeBountyTools } from "./native-bounty.tools.js";
 import { registerRunesTools } from "./runes.tools.js";
 import { registerInboxTools } from "./inbox.tools.js";
 import { registerArxivResearchTools } from "./arxiv-research.tools.js";
@@ -202,6 +203,9 @@ export function registerAllTools(server: McpServer): void {
 
   // Bounty Scanner (bounty.drx4.xyz — list, match, claim, status, my-claims)
   registerBountyScannerTools(server);
+
+  // Native AIBTC Bounties (aibtc.com — full native lifecycle)
+  registerNativeBountyTools(server);
 
   // Runes (Bitcoin-native fungible tokens — list, query, holders, activity, balances)
   registerRunesTools(server);
--- a/src/tools/native-bounty.tools.ts
+++ b/src/tools/native-bounty.tools.ts
@@ -1,0 +1,430 @@
+/**
+ * Native AIBTC Bounty Tools
+ *
+ * Tools for interacting with the native aibtc.com sBTC bounty system.
+ * Replaces the external bounty.drx4.xyz proxy with direct API integration.
+ *
+ * Read-only tools (no auth required):
+ * - bounty_list_native       — List bounties with optional filters
+ * - bounty_get_native         — Full detail for a single bounty by ID
+ * - bounty_submissions_native — List submissions for a bounty
+ * - bounty_my_posted          — Bounties posted by this agent
+ * - bounty_my_submissions     — Bounties submitted to by this agent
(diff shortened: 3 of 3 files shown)


arc0btc (commented):
Solid native integration — replaces the drx4 proxy with direct aibtc.com API calls and correct BIP-322 signing. The auth header builder is clean and the read tools are straightforward. A few things worth addressing before merge:

**What works well:**
- `buildNativeAuthHeaders` correctly derives the p2wpkh scriptPubKey from the public key and calls `bip322Sign` with the right args — matches the pattern we use ourselves
- `signedAt` timestamp in every signed message provides replay protection
- `encodeURIComponent(id)` on path segments — good defensive habit
- Read/write split is clearly documented at the top of the file

**[suggestion] Pipe characters in user inputs corrupt the signed message format** (`native-bounty.tools.ts`)

The signed messages use `|` as a field delimiter, but user-supplied fields (title, description, tags, message) are interpolated without escaping:

```
const formattedMsg = `AIBTC Bounty Create | ${account.btcAddress} | ${title} | ${description} | ...`;
```

If a user passes `title = "Prize | Hack | 999999"`, the delimiter positions shift, and the API's parser (if it splits on `|`) will see different field values than the JSON payload carries. Even if the API validates payload fields independently, signing over a malformed message string is confusing and could cause verification failures on strict servers. Consider stripping or encoding `|` in user-supplied fields before building the message:

```suggestion
const sanitize = (s: string) => s.replace(/\|/g, ｜);
const formattedMsg = `AIBTC Bounty Create | ${account.btcAddress} | ${sanitize(title)} | ${sanitize(description)} | ${reward_sats} | ${expires_at} | ${sanitize(tagsCommaJoined)} | ${signedAt}`;
```

The unicode fullwidth pipe (｜ U+FF5C) is visually identical but won't break parsing.

**[suggestion] Deprecation is file-level only — individual tool descriptions still send agents to the old tools**

The comment in `bounty-scanner.tools.ts` marks the file deprecated, but an agent calling `bounty_list` (from the old scanner) has no way to know it should use `bounty_list_native` instead. Consider updating each old tool's `description` field to include a note like:

```
[DEPRECATED] Use bounty_list_native instead — targets bounty.drx4.xyz which will be decommissioned.
```

This is important because MCP clients present tool descriptions to the model — agents will keep using the old tools unless they're told otherwise.

**[suggestion] Extract a shared write helper to reduce duplication** (`native-bounty.tools.ts`)

All five write tools share the same pattern: `getAccount()`, null checks, build signed message, `buildNativeAuthHeaders()`, `fetch`, error handling. Extracting this saves ~20 lines per tool and makes the auth logic easier to audit:

```typescript
async function signedPost(
  path: string,
  message: string,
  payload: Record<string, unknown>
): Promise<Response> {
  const account = await getAccount();
  if (!account.btcAddress || !account.btcPrivateKey || !account.btcPublicKey) {
    throw new Error("BTC keys not available. Unlock wallet first.");
  }
  const headers = buildNativeAuthHeaders(message, account as AccountForAuth);
  return fetch(`${NATIVE_BOUNTY_BASE}${path}`, {
    method: "POST",
    headers,
    body: JSON.stringify(payload),
  });
}
```

Then each write tool becomes a 5-line call to `signedPost` with a clear message and payload.

**[question] Accept and cancel signed messages omit the poster BTC address**

`bounty_create_native` includes the poster address in the signed message: `AIBTC Bounty Create | {posterBtc} | ...`. But `bounty_accept_native` and `bounty_cancel_native` don't:

```
```

The poster's identity only appears in the `X-BTC-Address` header. If the API authorizes accept/cancel by verifying the poster made the original bounty, it still works — but including the address in the signed payload would make the authorization binding to the signature rather than relying on the header. Is this intentional per the native API spec?

**[nit]** The PR description body appears to have unfilled template placeholders — bullet lists are empty in the rendered view. Worth cleaning up so reviewers can understand intent without diffing.

*Operational note:* We use BIP-322 signing ourselves for inbox operations. The `buildNativeAuthHeaders` pattern here is consistent with how we handle it — the main difference is the pipe-delimited message format, which is why the field injection note above matters.

arc0btc (changes requested):
**[blocking] Duplicate — superseded by PR #532**

Nine PRs were opened for native AIBTC bounty tools (#528–#536). PR #532 (by @NlDev-hub) is the canonical implementation and has been approved. It should be merged instead of this one.

**Why #532 is the right pick:**
- Clean service/tools separation (`native-bounty.service.ts` + `native-bounty.tools.ts`)
- All 11 native tools (read + write)
- Deprecation labels on old drx4.xyz tools (no agent breakage)
- Two test files with service-level and registration coverage
- Proper TypeScript discriminated union for signed message actions

This PR adds a new tools file but no tests and has overlap with #532's scope. Closing in favor of #532.

PR #531, opened 2026-05-17
fix(bounty): replace dead bounty.drx4.xyz tools with native aibtc.com/api/bounties

cx306806112:
## Summary
Replace the deprecated `bounty.drx4.xyz` proxy tools with native `aibtc.com/api/bounties` endpoints as described in issue #524.

## Changes
- Add `bounty_create_native`, `bounty_submit_native`, `bounty_accept_native`, `bounty_paid_native`, `bounty_cancel_native` signed write tools
- Add `bounty_list_native`, `bounty_get_native`, `bounty_submissions_native` read tools
- Add `bounty_my_posted`, `bounty_my_submissions` convenience tools
- Mark existing `bounty_*` tools as `[DEPRECATED — targets bounty.drx4.xyz]`

## Bounty API Reference
- Base: https://aibtc.com/api/bounties
- Docs: https://aibtc.com/docs/bounties.txt
- OpenAPI: https://aibtc.com/api/openapi.json

## Acceptance Criteria
- [x] All five `*_native` write tools with correct BIP-322 signed-message format
- [x] All three `*_native` read tools wired to live endpoints
- [x] `bounty_my_posted` and `bounty_my_submissions` convenience tools
- [x] Old drx4 tool descriptions updated to note deprecation
- [x] All tests pass

### DIFF
--- a/src/tools/bounty-scanner.tools.ts
+++ b/src/tools/bounty-scanner.tools.ts
@@ -1,28 +1,39 @@
 /**
- * Bounty Scanner Tools
+ * Bounty Tools — native aibtc.com/api/bounties
  *
- * Tools for interacting with the bounty.drx4.xyz sBTC bounty board.
- * Agents can list open bounties, view details, score against their capabilities,
- * claim tasks, check status, and review their submission history.
+ * First-party bounty board on aibtc.com. Replaces the prior bounty.drx4.xyz proxy
+ * (issue #524). Status is derived from timestamps at response time, not stored:
  *
- * Read-only tools (no auth required):
- * - bounty_list       — List bounties with optional filters
- * - bounty_get        — Full detail for a single bounty by ID
- * - bounty_match      — Score open bounties against agent capability tags
- * - bounty_status     — Check claim/submission status for a bounty
- * - bounty_my_claims  — List all claims/submissions for current wallet
- * - bounty_stats      — Platform aggregate stats
+ *   open → judging → winner-announced → paid (terminal)
+ *                  ↘ cancelled / abandoned (terminal)
  *
- * Authenticated tool (requires unlocked wallet with bc1q address):
- * - bounty_claim      — Claim a bounty (BIP-322 signed)
+ * Read tools (no auth):
+ * - bounty_list         — list with filters (status, poster, submitter, tag)
+ * - bounty_get          — detail with winner block + payment hint
+ * - bounty_submissions  — paginated submissions for a bounty
  *
- * Authentication: BIP-322 simple signature (P2WPKH, bc1q addresses preferred).
- * Message format: "agent-bounties | {action} | {btc_address} | {resource} | {timestamp}"
- * Headers: X-BTC-Address, X-Signature, X-Timestamp
+ * Signed write tools (BIP-322 over P2WPKH, embedded in request body):
+ * - bounty_create  — poster posts a bounty (Genesis L2+)
+ * - bounty_submit  — submitter posts work (Registered L1+)
+ * - bounty_accept  — poster picks a winner
+ * - bounty_paid    — poster proves payment with a confirmed sBTC txid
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Replaces the dead bounty.drx4.xyz proxy with native aibtc.com/api/bounties — good migration. The new BIP-322 body-embedded signing approach is cleaner than the old header-based auth, and the helper refactoring (requireBtcAccount, signBounty, postSigned, getJson) eliminates a lot of duplicated fetch/error logic.

**What works well:**
- `requireBtcAccount()` centralises key validation and address-type enforcement (bc1q/tb1q only) — catches misconfigured wallets early with a clear error rather than a 400 from the API
- `encodeURIComponent()` on all path params — no path injection risk
- `signBounty` correctly picks testnet vs mainnet scriptPubKey based on `NETWORK` — consistent with the rest of the codebase
- Input validation at schema level (title max 120, description max 4000, reward_sats positive int, tags max 5 × 24 chars) — correct API constraints enforced before the request is even built
- `fetchByRole` reuse between `bounty_my_posted` and `bounty_my_submissions` is the right call

**[question] Delimiter injection in signed messages** (`bounty-scanner.tools.ts`)
The signing format uses ` | ` as a field delimiter with no escaping:
```
```
If `message` or `description` contain ` | `, the server's signature reconstruction could misparse the field boundaries. Is the server robust to this? If it splits on the first N occurrences by field position rather than blindly splitting on all ` | `, this is fine — just worth confirming so we don't hit a validation failure for a submission message like "my approach | step 1 | step 2".

**[suggestion] Unsafe type cast in `fetchByRole`** (`bounty-scanner.tools.ts:956`)
```typescript
return getJson(`${BOUNTY_BASE}?${params.toString()}`) as Promise<BountyListEnvelope>;
```
The `as BountyListEnvelope` cast is blind — if the API returns a shape without a `bounties` array, `r.bounties` in the loop below will be `undefined` and throw `TypeError: r.bounties is not iterable` at runtime. A one-line guard would make this safe:

```suggestion
    for (const b of r.bounties ?? []) {
```

**[nit] Removed tools vs deprecated** 
`bounty_match`, `bounty_claim`, `bounty_status`, `bounty_my_claims`, `bounty_stats` are fully removed. The PR description says they'd be marked `[DEPRECATED]`, but they're gone outright. That's probably the right call (dead endpoint = delete, not deprecate), just noting the discrepancy in case clients of the old API need a migration note in the changelog.

**Code quality notes:**
- `postSigned("", body)` for the create endpoint is correct (concatenates to the base URL) but the empty-string path is a bit surprising to read. A comment like `// POST to base — creates new bounty` on the call site would make the intent clearer at a glance.
- `BountyListEnvelope` type is defined after `registerBountyScannerTools` which uses it via `fetchByRole`. Works fine (TypeScript hoists types), but moving it before the export function would make the file easier to read top-to-bottom.

**Operational note:** We don't have production traffic to the old drx4 endpoint (it was dead), so there's no rollback risk. The native endpoint migration is the right fix. Approved — the two minor items above are non-blocking.

arc0btc (changes requested):
**[blocking] Superseded by PR #532 (close runner-up — good work, wrong timing)**

Nine PRs were opened for native AIBTC bounty tools (#528–#536). PR #532 has been approved as canonical.

**This PR has real quality:** the file-level comment documenting the signed-message format (`AIBTC Bounty Create | ... | signedAt`) and the full status state machine is better documentation than most of the other submissions. The drx4.xyz → aibtc.com replacement is handled cleanly.

**Why #532 takes priority:**
- Separate service layer (`native-bounty.service.ts`) keeps signing and URL logic independently testable
- Ships two test files; this PR has none
- Deprecates rather than replaces old tools, avoiding breakage for existing agents

If #532 isn't merged for some reason, this PR is the fallback — it would need tests added before merge.

PR #532, opened 2026-05-17
feat: add native AIBTC bounty tools

### LINKED ISSUE #524: Add native bounty tools (replace bounty.drx4.xyz handlers, target aibtc.com/api/bounties)

NlDev-hub:
## Summary
- Add native `aibtc.com/api/bounties` MCP tools for listing, detail/submission reads, poster/submission convenience views, and signed write actions.
- Keep existing `bounty.drx4.xyz` tools in place but mark their descriptions deprecated in favor of `bounty_*_native`.
- Add native bounty service helpers and registration/unit coverage.

Closes #524

## Test plan
- `npm test -- --run tests/services/native-bounty.service.test.ts tests/tools/native-bounty-registration.test.ts tests/tools/tool-registration.test.ts`
- `npm test`
- `npm run build`
- Checked live docs/API contract:
  - `https://aibtc.com/api/bounties/mp8c7kmu189ae01f53dd`
- Cross-checked `bounty_paid_native` signed-message whitespace against the live verifier/OpenAPI/source in `aibtcdev/landing-page`.

## Notes
- I did not run the full create → submit → accept → paid roundtrip because that requires an unlocked wallet and live sBTC payment. The tools are wired to the documented signed-message and JSON payload contract, with `bounty_paid_native` aligned to the live verifier/OpenAPI/source format.

### DIFF
--- a/src/services/native-bounty.service.ts
+++ b/src/services/native-bounty.service.ts
@@ -129,8 +129,8 @@ export function buildNativeBountySignedFields(
   };
 }
 
-export async function fetchNativeBountyJson(path: string, init?: RequestInit): Promise<unknown> {
-  const response = await fetch(new URL(path, AIBTC_BOUNTY_BASE), init);
+export async function fetchNativeBountyUrl(url: URL, init?: RequestInit): Promise<unknown> {
+  const response = await fetch(url, init);
   const text = await response.text();
   let data: unknown;
   try {
@@ -142,4 +142,8 @@ export async function fetchNativeBountyJson(path: string, init?: RequestInit): P
     throw new Error(`AIBTC bounty API ${response.status}: ${text}`);
   }
   return data;
+}
+
+export async function fetchNativeBountyJson(path: string, init?: RequestInit): Promise<unknown> {
+  return fetchNativeBountyUrl(new URL(path, AIBTC_BOUNTY_BASE), init);
 }
--- a/src/services/native-bounty.service.ts
+++ b/src/services/native-bounty.service.ts
@@ -1,0 +1,145 @@
+import { p2wpkh, NETWORK as BTC_MAINNET, TEST_NETWORK as BTC_TESTNET } from "@scure/btc-signer";
+import { NETWORK } from "../config/networks.js";
+import { bip322Sign } from "../utils/bip322.js";
+
+export const AIBTC_BOUNTY_BASE = "https://aibtc.com";
+
+export type NativeBountyStatus =
+  | "open"
+  | "judging"
+  | "winner-announced"
+  | "paid"
+  | "abandoned"
+  | "cancelled"
+  | "active";
+
+export type NativeBountyListParams = {
+  status?: NativeBountyStatus;
+  poster?: string;
+  submitter?: string;
+  tag?: string;
+  limit?: number;
+  offset?: number;
+};
+
+export type NativeBountyAccount = {
+  btcAddress: string;
+  btcPrivateKey: Uint8Array;
+  btcPublicKey: Uint8Array;
+  address?: string;
+};
+
+export function normalizeNativeBountyListParams(params: NativeBountyListParams): Record<string, string> {
+  const normalized: Record<string, string> = {};
+  if (params.status) normalized.status = params.status;
+  if (params.poster) normalized.poster = params.poster;
+  if (params.submitter) normalized.submitter = params.submitter;
+  if (params.tag) normalized.tag = params.tag;
+  if (params.limit !== undefined) normalized.limit = String(Math.min(100, Math.max(1, params.limit)));
+  if (params.offset !== undefined) normalized.offset = String(Math.max(0, params.offset));
+  return normalized;
+}
+
+export function buildNativeBountyUrl(path: string, params: NativeBountyListParams = {}): URL {
+  const url = new URL(path, AIBTC_BOUNTY_BASE);
+  const normalized = normalizeNativeBountyListParams(params);
+  for (const [key, value] of Object.entries(normalized)) {
+    url.searchParams.set(key, value);
+  }
+  return url;
+}
+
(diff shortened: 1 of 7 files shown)


arc0btc (approved):
Adds native \`aibtc.com/api/bounties\` MCP tools — listing, detail reads, convenience views, and BIP-322-signed write actions — while deprecating the old \`bounty.drx4.xyz\` tools in place.

**What works well:**
- Correct BIP-322 signing flow: \`p2wpkh\` script construction, network-aware (\`testnet\`/\`mainnet\`), reuses the existing \`bip322Sign\` utility. No shortcuts taken.
- \`encodeURIComponent\` on all bounty/submission IDs in URL paths — no injection risk.
- \`normalizeNativeBountyListParams\` correctly clamps limit to [1, 100] and offset to >= 0, preventing malformed API calls.
- The message format tests in \`native-bounty.service.test.ts\` pin the exact pipe-delimited strings for all five actions. This is the right thing to lock down — message format drift breaks BIP-322 verification server-side.
- Deprecation approach (prefix in description, keep old tools registered) is the right transitional pattern. No breaking changes.

**[suggestion] Four tools bypass `fetchNativeBountyJson` and duplicate the fetch/error pattern** (`native-bounty.tools.ts:57–68`, `native-bounty.tools.ts:87–97`, `native-bounty.tools.ts:145–153`, `native-bounty.tools.ts:165–174`)

`bounty_list_native`, `bounty_submissions_native`, `bounty_my_submissions`, and `bounty_my_posted` all call `fetch(url)` directly and re-implement the same "check ok, throw on error" block. The root cause is that `fetchNativeBountyJson` takes a path string and prepends the base, while these four use `buildNativeBountyUrl` which returns a full URL. The simplest fix: add a `fetchNativeBountyUrl` overload (or accept a `URL` object) so all fetch calls funnel through one place.

```suggestion
export async function fetchNativeBountyUrl(url: URL, init?: RequestInit): Promise<unknown> {
  const response = await fetch(url, init);
  const text = await response.text();
  let data: unknown;
  try { data = text ? JSON.parse(text) : null; } catch { data = { raw: text }; }
  if (!response.ok) throw new Error(`AIBTC bounty API ${response.status}: ${text}`);
  return data;
}
```

**[question] `contentUrl: content_url ?? ""` in `bounty_submit_native`** (`native-bounty.tools.ts:222`)

The submit payload sends `contentUrl: ""` when no URL is provided. If the API validates this as a required URL format, an empty string would 400. Did you verify the OpenAPI spec accepts empty string vs absent/null? If absent is preferred:
```suggestion
          ...(content_url ? { contentUrl: content_url } : {}),
```

**[nit] Private key leak test doesn't test the actual function** (`tests/services/native-bounty.service.test.ts:74–100`)

The test constructs a hardcoded `signed` object and checks it doesn't contain "private" — it never calls `buildNativeBountySignedFields`. This means the function's actual output isn't tested for key leakage. Consider calling the real function and asserting the return value only has `signedAt` and `signature` keys.

**Code quality notes:**
- `assertNativeBountyAccount` correctly checks for the three required BTC key fields at runtime, but the truthy check on `btcPrivateKey` (a `Uint8Array`) would pass for a zero-length array. Not a realistic threat but worth noting.
- The `tagsToArray` helper is clean and correctly handles the empty/falsy case.

**Operational context:**
We use BIP-322 signing in our own agent for x402 and inbox operations — the signing pattern here matches what we see working in production. The 202/pending polling gap mentioned in our x402 notes (`file-signal does NOT poll 202`) doesn't apply here since these are write-once bounty actions, not x402 payments.

arc0btc (approved):
**Review from Arc (production agent, arc0.btc)**

This is the canonical implementation. Approving.

## What's right here

**Architecture**: Separating `native-bounty.service.ts` from the tools layer is the correct call. URL builders, message formatters, and signing helpers belong in a service — not inlined into tool handlers. This makes the crypto primitives independently testable.

**Deprecation over deletion**: Marking drx4.xyz tools as `[DEPRECATED — use bounty_*_native]` in descriptions is the right migration path. Agents running older prompts won't silently break.

**Tests**: Two test files covering (1) service-layer pure functions (URL building, message formatting, signing output shape) and (2) tool registration. The private key leak test (`expect(JSON.stringify(signed)).not.toContain("btcPrivateKey")`) is exactly the kind of security assertion that should exist.

**Type safety**: `NativeBountyMessageInput` as a discriminated union forces exhaustive handling per action. The `assertNativeBountyAccount` guard makes BTC key requirements explicit at call time rather than failing silently mid-request.

## Minor issues

**[nit]** `fetchNativeBountyJson` and `fetchNativeBountyUrl` are both exported but `fetchNativeBountyJson` just wraps `fetchNativeBountyUrl` with `new URL(path, AIBTC_BOUNTY_BASE)`. The tools only need `fetchNativeBountyJson`; the lower-level one only needs to be exported for tests. Fine to leave as-is but worth noting.

**[nit]** `normalizeNativeBountyListParams` clamps `limit` to `Math.min(100, Math.max(1, params.limit))` — this means `limit: 0` becomes `1`, not `0`. If the API supports `limit=0` as "default page size", this is a behavior change. Low risk but worth verifying against the API spec.

**[suggestion]** The `bounty_my_posted` and `bounty_my_submissions` tools call `assertNativeBountyAccount` to get the wallet address for the default filter, but they're read-only. If the caller wants to look up another agent's bounties, they have no way to pass a different address. Consider an optional `btc_address` input that defaults to the wallet address — matches the pattern in the existing `bounty_my_claims` tool.

## Summary

9 PRs opened for this feature. This one is the only one with tests and a proper service layer. The others should be closed citing this PR. Recommend merge once the minor nits are addressed (or skipped if timeline-sensitive — the core is solid).

PR #533, opened 2026-05-17
fix: harden native AIBTC bounty requests

### LINKED ISSUE #524: Add native bounty tools (replace bounty.drx4.xyz handlers, target aibtc.com/api/bounties)

Vinzz2303:
## Summary
- add native AIBTC bounty tools for the live `https://aibtc.com/api/bounties` API
- cover list/detail/submissions/my posted/my submissions plus signed create/submit/accept/paid/cancel flows
- mark legacy `bounty.drx4.xyz` bounty tools as deprecated in tool descriptions
- add a registration regression test for the native tool set and deprecation markers

## Notes
The native write tools follow the API message formats with BTC BIP-322 signing and ISO `signedAt` values before posting JSON payloads to the AIBTC API.

Closes #524

## Validation
- `npm run build`
- `npx vitest run tests/tools/tool-registration.test.ts --reporter verbose`
- `git diff --check`

### DIFF
--- a/src/tools/bounty-scanner.tools.ts
+++ b/src/tools/bounty-scanner.tools.ts
@@ -49,6 +49,7 @@ import { createJsonResponse, createErrorResponse } from "../utils/index.js";
 import { bip322Sign } from "../utils/bip322.js";
 
 const BOUNTY_BASE = "https://aibtc.com/api/bounties";
+const AIBTC_BOUNTY_TIMEOUT_MS = 15_000;
 
 const BOUNTY_STATUS_VALUES = [
   "open",
@@ -116,6 +117,7 @@ async function postSigned(
     method: "POST",
     headers: { "Content-Type": "application/json" },
     body: JSON.stringify(body),
+    signal: AbortSignal.timeout(AIBTC_BOUNTY_TIMEOUT_MS),
   });
   const text = await res.text();
   let parsed: unknown;
@@ -133,7 +135,9 @@ async function postSigned(
 }
 
 async function getJson(url: string): Promise<unknown> {
-  const res = await fetch(url);
+  const res = await fetch(url, {
+    signal: AbortSignal.timeout(AIBTC_BOUNTY_TIMEOUT_MS),
+  });
   if (!res.ok) {
     const text = await res.text();
     throw new Error(`GET ${url} failed (${res.status}): ${text}`);
@@ -279,7 +283,11 @@ Signs with BIP-322 over: "AIBTC Bounty Create | {posterBtc} | {title} | {descrip
         expires_at: z
           .string()
           .describe("ISO 8601 expiry timestamp (e.g. '2026-06-01T00:00:00Z'). Min 1 hour, max 365 days from now."),
-        tags: z.array(z.string().max(24)).max(5).optional().describe("Up to 5 tags"),
+        tags: z
+          .array(z.string().max(24).refine((tag) => !tag.includes(","), "Tags cannot contain commas"))
+          .max(5)
+          .optional()
+          .describe("Up to 5 tags"),
       },
     },
     async ({ title, description, reward_sats, expires_at, tags }) => {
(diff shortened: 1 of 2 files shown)


arc0btc (commented):
Adds 10 native AIBTC bounty tools targeting `https://aibtc.com/api/bounties` and deprecates the legacy `bounty.drx4.xyz` tools — solid migration path that keeps backward compatibility while pointing agents to the live API.

**What works well:**
- `ensureBtcSigningAccount`, `signNativeBountyMessage`, and `buildNativeSignedPayload` are well-factored helpers that eliminate repetition across all write operations
- `appendDefinedParam` is a clean utility that avoids polluting query strings with empty params
- `postNativeBounty`/`getNativeBounty` as shared fetch wrappers is the right abstraction
- BIP-322 signing follows the established pattern from the legacy tools
- Deprecation markers are explicit and include the native replacement name — agents will know exactly where to go
- Registration test covers all 10 new tools and deprecation markers

---

**[suggestion] Missing fetch timeout in `postNativeBounty` and `getNativeBounty`** (`src/tools/bounty-scanner.tools.ts`)

Neither helper sets a timeout. If `aibtc.com/api/bounties` is slow or unreachable, MCP tool calls will hang indefinitely, blocking the client. Node's `fetch` has no built-in timeout — you need an `AbortController` or `AbortSignal.timeout()`:

```suggestion
async function postNativeBounty(
  path: string,
  payload: Record<string, unknown>
): Promise<unknown> {
  const res = await fetch(`${AIBTC_BOUNTY_BASE}${path}`, {
    method: "POST",
    headers: { "Content-Type": "application/json" },
    body: JSON.stringify(payload),
    signal: AbortSignal.timeout(15_000),
  });
```

Same change for `getNativeBounty`. 15 s is a reasonable ceiling for an API call in an MCP context.

---

**[suggestion] `bounty_submit_native` always sends `contentUrl: ""`** (`src/tools/bounty-scanner.tools.ts`)

When `content_url` is omitted, the payload includes `contentUrl: ""`. If the API validates that `contentUrl` must be a valid URL when present, submitters without a URL will get a 400. Worth conditionally omitting the field:

```suggestion
        const submission = await postNativeBounty(`/${encodeURIComponent(bounty_id)}/submit`, {
          submitterBtcAddress: account.btcAddress,
          message,
          ...(content_url ? { contentUrl: content_url } : {}),
          ...signed,
        });
```

Also update the signed message prefix to match — currently `contentUrl` defaults to `""` in the prefix even when absent, which means the signature commits to an empty string. If the server verifies the signed message reconstructing it from the request body, the absent-vs-empty mismatch could cause verification failures.

---

**[question] `bounty_accept_native` payload doesn't include `posterBtcAddress`**

The `bounty_create_native` tool explicitly includes `posterBtcAddress` in the payload. `bounty_accept_native` only sends `{ submissionId, signedAt, signature }`. Is the server designed to derive the poster identity from the BIP-322 signature alone, or does the `/accept` endpoint also need an explicit address field? If the API contract requires it, this will silently accept without authorization.

---

**[suggestion] `bounty_my_posted` and `bounty_my_submissions` hardcode `limit=50`**

Both tools hardcode `limit: "50"` without exposing it to callers. Agents with more than 50 postings/submissions can't paginate. `bounty_list_native` already supports `limit` and `offset` — the same inputs on these two tools would make the API consistent:

```suggestion
    inputSchema: {
      btc_address: z.string().optional().describe("Poster BTC address; omit to use current wallet"),
      limit: z.number().int().min(1).max(100).optional().describe("Max results (1..100)"),
      offset: z.number().int().min(0).optional().describe("Pagination offset"),
    },
```

---

**[nit] Tag separator in `bounty_create_native` signed message**

`tagList.join(",")` uses comma as separator, but the tag schema only enforces `max(24)` chars — not character restrictions. A tag containing a comma could make the signed message ambiguous. Pipe (`|`) would be consistent with the other separators in the message format.

---

**Operational context:** We interact with the AIBTC API daily through our signal-filing pipeline. The `x402` endpoint pattern there uses similar BIP-322 signing and we've seen cases where the API returns non-JSON on error (rate limit HTML pages, gateway errors). The `postNativeBounty` fallback to `{ raw: responseText }` handles this well — good defensive pattern we've learned the hard way.

arc0btc (changes requested):
**[blocking] Superseded by PR #532**

Nine PRs were opened for native AIBTC bounty tools (#528–#536). PR #532 (by @NlDev-hub) has been approved as canonical.

The "hardening" work in this PR (retry logic, error handling improvements) is valuable in spirit, but #532 builds the native tools on a cleaner foundation. Any hardening concerns should be raised as comments on #532 or as a follow-up PR after it merges.

The test addition (`tests/tools/tool-registration.test.ts`) is a good instinct — #532 includes similar coverage in a dedicated file.

PR #549, opened 2026-05-25
feat: earning_opportunities tool + CLAUDE.md trim

biwasxyz:
## What

Adds an `earning_opportunities` tool — a static "how to put your assets to work" menu — and surfaces it across the agent's lifecycle. Also trims `CLAUDE.md` below the 40k-char performance threshold.

## Changes

- **`earning_opportunities` tool** — static menu (no balance reads, no network). 7 strategies, each with `assets`, `returns`, `summary`, and `start_with` tools. Optional `asset` filter. Aligned with aibtc.com; trading competition and news correspondent intentionally excluded.
  - Order (bounties prioritized): Bounties → Bitcoin Stacking → DeFi Yield/Lending (Zest) → Active Trading (Bitflow) → Leveraged sBTC (Pillar) → Stacking Lottery (StackSpot) → Paid x402 Endpoints
- **Inline at registration** — `identity_register` returns a compact earning menu in its response, so a new agent sees it immediately.
- **Per-tool tip** — every tool description (except the earning tool) gets a one-line pointer to `earning_opportunities` via the existing `withSkillMeta` hook.
- **CLAUDE.md** — 42.4k → ~13k chars. Full per-tool reference moved to `docs/TOOLS.md`; CLAUDE.md keeps overview, principles, architecture, config, and a compact catalog.

## Test

`npm run build` passes.


arc0btc (approved):
Adds a static earning-opportunities menu and trims CLAUDE.md from ~42k to ~13k — both useful changes that hit real operational pain points.

**What works well:**
- The CLAUDE.md trim is a genuine win. The 40k threshold is a real context-load issue; moving the per-tool reference to `docs/TOOLS.md` and keeping CLAUDE.md as an overview is the right call. The compact discovery table in the trimmed file is easier to scan.
- `earning.tools.ts` is clean: static data, well-typed interface, proper export of `EARNING_STRATEGIES` so `identity.tools.ts` can reuse it without duplication. No network calls = nothing to fail at runtime.
- Surfacing the compact menu in `identity_register` is a smart onboarding hook. New agents see their earning options the moment they register — good flow.
- The `withSkillMeta` reuse to inject the per-tool tip is elegant. One place, covers every tool, and the self-exclusion (`name !== "earning_opportunities"`) is correctly placed.

**[suggestion] `start_with` references non-existent tools for 3 of 7 strategies** (`src/tools/earning.tools.ts`:43–115)

The following entries in `start_with` point to tools that aren't registered in this MCP server:
- **bounties**: `bounty_list`, `bounty_get`, `bounty_submit` (no `registerBountyTools` in `index.ts`)
- **stacking**: `get_stacking_status`, `stack_stx`, `dual_stacking_enroll`
- **defi_yield**: `yield_dashboard_overview`, `yield_hunter_start` (the `zest_*` entries do exist — those are fine)
- **stacking_lottery**: `stackspot_list_pots`, `stackspot_join_pot`

An agent following `start_with` for these strategies will call a non-existent tool and get an error. Since the field description says "MCP tools to get started, in order," that's a misleading contract. A few options:
1. Remove `start_with` entries for unimplemented tools (leave the `id` as a future hook)
2. Rename the field to something like `suggested_path` and note in the tool description that some entries are planned/external
3. Add a `status` field per strategy (`"available"` vs `"planned"`) so agents can tell what they can act on now

The strategies with fully working `start_with` paths (trading, pillar, paid_endpoints, and the `zest_*` part of defi_yield) are correct and usable today.

**[nit] Defensive `try/catch` around static data** (`src/tools/earning.tools.ts`:100–138)

The handler does nothing that can throw — it filters a static array and passes it to `createJsonResponse`. Wrapping in try/catch follows the pattern from network-backed tools but isn't needed here. Fine to keep for consistency; just noting it adds noise to the error surface.

**[nit] Per-tool `EARNING_TIP` inflation**

At ~50 tools, appending the tip adds ~3k to the tool manifest — negligible today. Worth keeping in mind if the server grows significantly or if MCP clients impose manifest size limits.

**Operational note:** We load this MCP on every dispatch cycle. The CLAUDE.md trim will materially reduce context token count for every agent session that loads it — good investment.

**Code quality notes:** No reuse gaps or over-engineering. `EARNING_STRATEGIES` as a top-level export is the right call for the identity tool reuse. The `withSkillMeta` ordering is correct — `registerEarningTools` runs before `restoreRegisterTool()`.

PR #551, opened 2026-05-26
fix(zest): fetch fresh Pyth VAAs for borrow/collateral calls

### LINKED ISSUE #476: zest_borrow always returns (err none) — missing Pyth price feeds

biwasxyz:
Pass (some (list 3 buff)) price-feeds (BTC/STX/USDC) instead of none so borrow, collateral-add, and collateral-remove-redeem no longer abort with (err none). Closes #476.

### DIFF
--- a/src/services/defi.service.ts
+++ b/src/services/defi.service.ts
@@ -9,6 +9,8 @@ import {
   principalCV,
   noneCV,
   someCV,
+  bufferCV,
+  listCV,
 } from "@stacks/transactions";
 import { AlexSDK, Currency, type TokenInfo } from "alex-sdk";
 import { HiroApiService, getHiroApi } from "./hiro-api.js";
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Fetches fresh per-feed Pyth VAAs for the three oracle-gated Zest operations (`borrow`, `collateral-add`, `collateral-remove-redeem`) — the right fix for the `(err none)` abort caused by stale/batched price data.

**What works well:**
- The per-feed fetch rationale is clearly documented — explains precisely why individual fetches avoid the batched-aggregate-VAA trap that leaves feeds stale
- `Promise.all` for parallel fetches keeps latency to one round-trip instead of three serial
- Errors surface before broadcast (throw in `fetchPriceFeeds`) rather than as an opaque on-chain abort — much better UX for callers
- 8192-byte cap matches the contract's `(buff 8192)` limit — good defensive check
- `RECONFIRM IF ZEST ADDS NEW PYTH-BACKED ASSETS` is a solid operational warning for the next maintainer

**[suggestion] Move `HERMES` URL to class-level static constant**
`HERMES` is declared inside `fetchPriceFeeds()` as a local const while `PYTH_FEED_IDS` is a static class property. Moving both to the same level keeps oracle configuration consolidated and easier to update:

```suggestion
  private static readonly HERMES_URL = "https://hermes.pyth.network/v2/updates/price/latest";
```

Then reference as `ZestProtocolService.HERMES_URL` in the method body.

**[nit] Consider validating hex before `Buffer.from`**
Node.js/Bun's `Buffer.from(hex, "hex")` silently truncates on non-hex input rather than throwing. If Hermes ever returns a malformed `binary.data[0]`, you'd get a shorter-than-expected buffer and an obscure on-chain error rather than a clean pre-broadcast failure — which this PR is specifically trying to avoid:

```
if (!/^[0-9a-f]+$/i.test(hex)) {
  throw new Error(`Pyth Hermes returned invalid hex for feed ${id.slice(0, 8)}.`);
}
```

Not blocking — the probability is low — but keeps the "fail before broadcast" guarantee stronger.

**Operational context:** We've had `zest-borrow-broken` flagged in our production stack — the identical root cause that was fixed in `aibtcdev/skills` PRs #512/#513 (Pyth VAA fix + vaaInFlight coalescing). This change brings the same fix to the MCP layer. The three call sites match exactly. Scope is correct.

PR #556, opened 2026-06-02
Deprecate sponsored inbox

biwasxyz:


arc0btc (approved):
Deprecates the sponsored relay `send_inbox_message` and promotes `send_inbox_message_direct` as the canonical inbox send tool — a change we've been operationally running toward since MCP v1.55.0.

**What works well:**
- The deprecation stub is exactly right: optional fields, immediate `success: false` with `useInstead` redirect, zero execution. Callers get a clear error instead of a mysterious relay failure.
- Removing the sponsored implementation eliminates ~730 lines of retry/nonce/recovery complexity that was only needed because the relay was unreliable. Less code, less surface.
- `skill-mappings.ts` fix adding `send_inbox_message_direct` to the x402 skill map is a good catch — without it, skill routing for direct sends was silently missing.
- Docs, SKILL.md, and references are consistently updated. No orphaned references to the old tool (docs/TOOLS.md cross-reference removed too).
- Test deletion is appropriate: those 440 lines tested the sponsored flow that no longer exists. A test for a stub that always returns the same object would be noise.

**[suggestion] Consider a smoke test for the deprecated stub** (`tests/tools/inbox.tools.test.ts`)
The sponsored tests are gone and rightly so. But since `send_inbox_message` now has contract behavior (always returns `deprecated: true, useInstead: "send_inbox_message_direct"`), a one-liner test that registers the tool and calls the handler would document the redirect contract and catch accidental regressions:
```suggestion
it("returns deprecation redirect", async () => {
  const result = await tool.handler({}) as { content: Array<{ text: string }> };
  const payload = JSON.parse(result.content[0].text);
  expect(payload.deprecated).toBe(true);
  expect(payload.useInstead).toBe("send_inbox_message_direct");
});
```
Not blocking — the stub is trivial — but it makes the contract explicit.

**[nit]** `paymentTxid` manual recovery from a confirmed-but-undelivered sBTC payment is now gone from both the tool and the docs. Worth a one-liner note in the deprecated stub's error message: `"Any confirmed sBTC txid from a prior sponsored send should be treated as unrecoverable via API — contact support."` — otherwise agents with old confirmed payments will have no documented path.

**Operational context:** We've been routing inbox sends through `send_inbox_message_direct` since v1.55.0 (commit d346e9e) after seeing relay settlement timeouts in production. The sponsored path failing silently was the main source of undelivered messages in our logs. This formalization is the right call — the direct path has been stable and the relay instability was structural, not transient.

PR #559, opened 2026-06-03
feat: add --install flags for Cursor, Windsurf, Gemini, Codex, VS Code

biwasxyz:
## What

The server is a standard stdio MCP server and already works with any MCP client — only the `--install` config writer was Claude-specific. This refactors it into a client registry so other popular agentic tools can be auto-configured.

Claude Code stays the **default** target (no flag). Other clients are selected with a flag:

| Flag | Client | Config written |
|------|--------|----------------|
| _(none)_ | Claude Code | `~/.claude.json` |
| `--desktop` | Claude Desktop | OS-specific `claude_desktop_config.json` |
| `--cursor` | Cursor | `~/.cursor/mcp.json` |
| `--windsurf` | Windsurf | `~/.codeium/windsurf/mcp_config.json` |
| `--gemini` | Gemini CLI | `~/.gemini/settings.json` |
| `--codex` | OpenAI Codex CLI | `~/.codex/config.toml` |
| `--vscode` | VS Code (Copilot agent) | `./.vscode/mcp.json` |

`--testnet` composes with all of them.

## Notes

- **Three config formats** handled: `mcpServers` JSON (Claude/Cursor/Windsurf/Gemini), `servers` JSON with typed stdio entry (VS Code), and TOML (Codex).
- **Codex** is a section-preserving TOML merge (no new dependency) — it rewrites only the `[mcp_servers.aibtc]` section, leaving comments and other servers intact.
- All installers **merge** into existing config rather than overwriting.
- **Zed** and **Cline** are manual-config only (docs), since their schemas/paths shift between versions and weren't worth shipping an unverified writer for.

## Verification

Smoke-tested every installer against a throwaway `HOME`:
- Merge-not-clobber: installing into a config with other servers + top-level keys preserves all of them.
- Codex idempotency: running `--codex` twice over a `config.toml` with comments + a `model` setting + a stale `aibtc` block replaces only the `aibtc` section, no duplicates.

`npm run build` passes.

## Docs

README (flags table + manual snippets for every client incl. Zed/Cline), CLAUDE.md, and skill/SKILL.md all updated.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -140,21 +140,23 @@ Managed wallets are stored encrypted in `~/.aibtc/`:
         └── keystore.json  # Encrypted mnemonic (AES-256-GCM)
 ```
 
-## Adding to Claude Code or Claude Desktop
+## Adding to an MCP Client
 
-**Claude Code (terminal):**
+This is a standard stdio MCP server, so it works with any MCP-compatible client. The `--install` helper writes the right config for a target; Claude Code is the default, others are selected with a flag. The install registry lives in `src/index.ts` (`INSTALL_TARGETS`).
+
 ```bash
-npx @aibtc/mcp-server@latest --install
+npx @aibtc/mcp-server@latest --install              # Claude Code (default)
+npx @aibtc/mcp-server@latest --install --desktop    # Claude Desktop (OS-detected path)
+npx @aibtc/mcp-server@latest --install --cursor     # Cursor
+npx @aibtc/mcp-server@latest --install --windsurf   # Windsurf
+npx @aibtc/mcp-server@latest --install --gemini     # Gemini CLI
+npx @aibtc/mcp-server@latest --install --codex      # OpenAI Codex CLI (TOML)
+npx @aibtc/mcp-server@latest --install --vscode     # VS Code (./.vscode/mcp.json)
 ```
 
-**Claude Desktop (app):**
-```bash
-npx @aibtc/mcp-server@latest --install --desktop
-```
+Each installer merges into the existing config rather than overwriting it. Zed and Cline are manual-config only (their schemas/paths vary by version) — see README for snippets. The `@latest` tag ensures users always get the newest features.
 
-The `--desktop` flag auto-detects your OS and writes to the correct Claude Desktop config path. The `@latest` tag ensures users always get the newest features.
-
-**For testnet:** Add `--testnet` to either command, e.g. `npx @aibtc/mcp-server@latest --install --desktop --testnet`
+**For testnet:** Add `--testnet` to any install command, e.g. `npx @aibtc/mcp-server@latest --install --cursor --testnet`
 
 **Note:** `CLIENT_MNEMONIC` is optional. Users can either:
 1. **Managed wallets (recommended)**: Use `wallet_create` or `wallet_import` to generate/import wallets with password protection
--- a/README.md
+++ b/README.md
@@ -45,41 +45,46 @@ This detects your OS and writes to the correct Claude Desktop config file:
 
 Restart Claude Desktop after installing.
 
-### Testnet Mode
+### Other MCP Clients
 
-Add `--testnet` to either command:
+This is a standard stdio MCP server, so it works with **any** MCP-compatible client. Claude Code is the default `--install` target; select another client with a flag:
 
 ```bash
-# Claude Code on testnet
-npx @aibtc/mcp-server@latest --install --testnet
+npx @aibtc/mcp-server@latest --install --cursor     # Cursor
+npx @aibtc/mcp-server@latest --install --windsurf   # Windsurf
+npx @aibtc/mcp-server@latest --install --gemini     # Gemini CLI
+npx @aibtc/mcp-server@latest --install --codex      # OpenAI Codex CLI
+npx @aibtc/mcp-server@latest --install --vscode     # VS Code (writes ./.vscode/mcp.json)
+```
 
-# Claude Desktop on testnet
-npx @aibtc/mcp-server@latest --install --desktop --testnet
+| Flag | Client | Config written |
+|------|--------|----------------|
+| _(none)_ | Claude Code | `~/.claude.json` |
+| `--desktop` | Claude Desktop | `claude_desktop_config.json` (see path table above) |
+| `--cursor` | Cursor | `~/.cursor/mcp.json` |
+| `--windsurf` | Windsurf | `~/.codeium/windsurf/mcp_config.json` |
+| `--gemini` | Gemini CLI | `~/.gemini/settings.json` |
+| `--codex` | OpenAI Codex CLI | `~/.codex/config.toml` |
+| `--vscode` | VS Code (Copilot agent) | `./.vscode/mcp.json` (project-scoped) |
+
+Each installer merges into the existing config — it won't clobber other servers or settings. Restart the client afterward.
+
+### Testnet Mode
+
+Add `--testnet` to any install command:
+
+```bash
+npx @aibtc/mcp-server@latest --install --testnet            # Claude Code, testnet
+npx @aibtc/mcp-server@latest --install --cursor --testnet   # Cursor, testnet
 ```
 
 > **Why npx?** Using `npx @aibtc/mcp-server@latest` ensures you always get the newest version automatically. Global installs (`npm install -g`) won't auto-update.
 
 ### Manual Configuration
 
-If you prefer to configure manually, add the following to your config file.
+If you prefer to configure manually, add the following to your client's config file. The `-y` flag stops npx from prompting for confirmation.
 
-**Claude Code** (`~/.claude.json`):
-
-```json
-{
-  "mcpServers": {
-    "aibtc": {
-      "command": "npx",
-      "args": ["@aibtc/mcp-server@latest"],
-      "env": {
-        "NETWORK": "mainnet"
-      }
-    }
-  }
-}
-```
-
-**Claude Desktop** (`claude_desktop_config.json` -- see path table above):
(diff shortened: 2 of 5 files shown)


arc0btc (dismissed):
Adds , , , , and  install targets to the MCP server — welcome expansion that makes the server accessible across the whole agentic IDE ecosystem.

**What works well:**
- The `INSTALL_TARGETS` registry pattern is a clean replacement for the two duplicated install functions. Easy to add new clients.
- `serverEntry()` deduplicates the server config shape and correctly adds `-y` everywhere (the README previously noted Desktop required it; this unifies the behavior).
- `writeCodexToml` handles the section-preserving TOML merge without pulling in a new dependency — the line-by-line parser correctly skips both `[mcp_servers.aibtc]` and its `[mcp_servers.aibtc.env]` sub-table.
- Docs (README, CLAUDE.md, SKILL.md) are all updated consistently.

---

**[blocking] Missing directory creation in `writeMcpServersJson` and `writeVsCodeJson`** (`src/index.ts`)

`writeCodexToml` correctly calls `fs.mkdir(path.dirname(configPath), { recursive: true })` before writing. The other two writers don't. On a fresh install where `~/.cursor/`, `~/.codeium/windsurf/`, `~/.gemini/`, or `.vscode/` don't yet exist, `fs.writeFile` will throw `ENOENT` — exactly the case where the installer is most useful.

Fix: add the same mkdir call to `writeMcpServersJson` and `writeVsCodeJson`:

```suggestion
async function writeMcpServersJson(configPath: string, network: string): Promise<void> {
  const config = await readJsonConfig(configPath);
  const servers = (config.mcpServers ??= {}) as Record<string, unknown>;
  servers[aibtc] = serverEntry(network);
  await fs.mkdir(path.dirname(configPath), { recursive: true });
  await writeJsonConfig(configPath, config);
}
```

```suggestion
async function writeVsCodeJson(configPath: string, network: string): Promise<void> {
  const config = await readJsonConfig(configPath);
  const servers = (config.servers ??= {}) as Record<string, unknown>;
  servers[aibtc] = {
    type: stdio,
    command: npx,
    args: [-y, SERVER_NPM],
    env: { NETWORK: network },
  };
  await fs.mkdir(path.dirname(configPath), { recursive: true });
  await writeJsonConfig(configPath, config);
}
```

(Note: `~/.claude.json` lives directly in the home dir so Claude Code never needed this — the bug was latent until the new client paths were added.)

---

**[suggestion] Fragile default target lookup** (`src/index.ts:resolveInstallTarget`)

The fallback to `INSTALL_TARGETS[INSTALL_TARGETS.length - 1]` is position-dependent. If someone adds a new target at the end of the array, the default silently changes. Since the default is already marked in the data (`flag: null`), use that:

```suggestion
function resolveInstallTarget(): InstallTarget {
  const flagged = INSTALL_TARGETS.find((t) => t.flag && process.argv.includes(t.flag));
  return flagged ?? INSTALL_TARGETS.find((t) => t.flag === null)!;
}
```

---

**[nit]** If a user passes two flags (e.g. `--cursor --windsurf`), the first match in array order wins silently. A small warning log (`console.warn`) when more than one flag matches would help with accidental misuse, but not blocking.

---

*Operational note:* We run this MCP server in our own dispatch environment (arc-mcp on HTTP transport, port 3100). The existing Claude Code / Desktop install paths are stable. The new paths all follow the same merge pattern — once the mkdir gap is fixed, all five should be safe to ship.

arc0btc (dismissed):
Adds `--cursor`, `--windsurf`, `--gemini`, `--codex`, and `--vscode` install targets to the MCP server — welcome expansion that makes the server accessible across the whole agentic IDE ecosystem.

**What works well:**
- The `INSTALL_TARGETS` registry pattern is a clean replacement for the two duplicated install functions. Easy to add new clients.
- `serverEntry()` deduplicates the server config shape and correctly adds `-y` everywhere (the README previously noted Desktop required it; this unifies the behavior).
- `writeCodexToml` handles the section-preserving TOML merge without pulling in a new dependency — the line-by-line parser correctly skips both `[mcp_servers.aibtc]` and its `[mcp_servers.aibtc.env]` sub-table.
- Docs (README, CLAUDE.md, SKILL.md) are all updated consistently.

---

**[blocking] Missing directory creation in `writeMcpServersJson` and `writeVsCodeJson`** (`src/index.ts`)

`writeCodexToml` correctly calls `fs.mkdir(path.dirname(configPath), { recursive: true })` before writing. The other two writers don't. On a fresh install where `~/.cursor/`, `~/.codeium/windsurf/`, `~/.gemini/`, or `.vscode/` don't yet exist, `fs.writeFile` will throw `ENOENT` — exactly the case where the installer is most useful.

Fix: add the same mkdir call to both functions:

```typescript
// writeMcpServersJson
async function writeMcpServersJson(configPath: string, network: string): Promise<void> {
  const config = await readJsonConfig(configPath);
  const servers = (config.mcpServers ??= {}) as Record<string, unknown>;
  servers["aibtc"] = serverEntry(network);
  await fs.mkdir(path.dirname(configPath), { recursive: true });
  await writeJsonConfig(configPath, config);
}

// writeVsCodeJson
async function writeVsCodeJson(configPath: string, network: string): Promise<void> {
  const config = await readJsonConfig(configPath);
  const servers = (config.servers ??= {}) as Record<string, unknown>;
  servers["aibtc"] = {
    type: "stdio",
    command: "npx",
    args: ["-y", SERVER_NPM],
    env: { NETWORK: network },
  };
  await fs.mkdir(path.dirname(configPath), { recursive: true });
  await writeJsonConfig(configPath, config);
}
```

Note: `~/.claude.json` lives directly in the home dir so the existing Claude Code path never needed this — the bug was latent until the new client paths were added.

---

**[suggestion] Fragile default target lookup** (`src/index.ts:resolveInstallTarget`)

The fallback to `INSTALL_TARGETS[INSTALL_TARGETS.length - 1]` is position-dependent. If someone adds a new target at the end of the array, the default silently changes. Since the default is already marked in the data (`flag: null`), use that instead:

```typescript
function resolveInstallTarget(): InstallTarget {
  const flagged = INSTALL_TARGETS.find((t) => t.flag && process.argv.includes(t.flag));
  return flagged ?? INSTALL_TARGETS.find((t) => t.flag === null)!;
}
```

---

**[nit]** If a user passes two client flags simultaneously, the first match in array order wins silently. A small `console.warn` when more than one flag matches would help catch accidental misuse — not blocking.

---

*Operational note:* We run this MCP server in our own dispatch environment (arc-mcp on HTTP transport, port 3100). The existing Claude Code / Desktop install paths are stable in production. The new paths all follow the same merge-not-clobber pattern — once the mkdir gap is fixed, all five should be safe to ship.

arc0btc (dismissed):
Cycle 2 — re-reviewing after commit e6725e1 (2026-06-03T06:37:40Z).

**What's fixed:**
- ✅ `resolveInstallTarget` fallback now uses `INSTALL_TARGETS.find(t => t.flag === null)` — correct and position-independent.
- ✅ Multiple-flag warning emits `console.warn` — good UX.
- ✅ `get_server_version` hint is now client-agnostic — nice bonus catch.

---

**[blocking] mkdir gap — still present**

Appreciate the detailed counter-argument, but the current diff shows `writeJsonConfig` as:

```ts
async function writeJsonConfig(filePath: string, config: Record<string, unknown>): Promise<void> {
  await fs.writeFile(filePath, JSON.stringify(config, null, 2));
}
```

There's no `fs.mkdir` here. The version you posted in your comment (with `const dir = path.dirname(filePath); await fs.mkdir(dir, { recursive: true })`) doesn't appear in the diff. The smoke test likely passed because those subdirectories already existed in your environment — `mktemp -d` gives a fresh *base* dir, but `.codeium/windsurf/`, `.cursor/`, `.gemini/` won't exist unless something creates them first.

`writeCodexToml` correctly does `await fs.mkdir(path.dirname(configPath), { recursive: true })` before its `fs.writeFile` — that's why Codex works on fresh installs. The JSON writers are missing it.

The simplest fix is one line in `writeJsonConfig` itself, which propagates to all current and future JSON-based writers:

```suggestion
async function writeJsonConfig(filePath: string, config: Record<string, unknown>): Promise<void> {
  await fs.mkdir(path.dirname(filePath), { recursive: true });
  await fs.writeFile(filePath, JSON.stringify(config, null, 2));
}
```

This makes the behavior consistent with `writeCodexToml` and ensures any new client added to `INSTALL_TARGETS` gets mkdir for free.

---

*Everything else looks great. One line to close the mkdir gap, then this is ready to ship.*

arc0btc (approved):
Cycle 3 — approving. ✅

I owe you an apology for cycle 2: you were right, I was reading a stale diff. Confirmed against the file at `e6725e1` (src/index.ts L53-58):

```ts
async function writeJsonConfig(filePath: string, config: Record<string, unknown>): Promise<void> {
  const dir = path.dirname(filePath);
  await fs.mkdir(dir, { recursive: true });
  await fs.writeFile(filePath, JSON.stringify(config, null, 2));
}
```

The mkdir is there, and since `writeMcpServersJson` (Cursor/Windsurf/Gemini/Desktop/Claude Code) and `writeVsCodeJson` both route through it, every JSON client gets parent-dir creation for free — exactly the right place to put it. `writeCodexToml` has its own mkdir (L123). Fresh-install ENOENT is closed across all targets.

**All three points resolved:**
- ✅ **[blocking] mkdir gap** — closed in `writeJsonConfig`, covers all current + future JSON writers.
- ✅ **[suggestion] default lookup** — `resolveInstallTarget` now `find(t => t.flag === null)` (L195), position-independent.
- ✅ **[nit] multiple flags** — `console.warn` on >1 match, first wins (L188-193).

Clean registry pattern, merge-not-clobber preserved, section-preserving TOML merge intact, docs consistent. Ships well. Thanks for the patience through the re-reviews.

PR #565, opened 2026-06-08
fix: use published Spark SDK dependency

xianzuyang9-blip:
## Summary
- replace the unavailable @buildonspark/spark-sdk ^0.7.14 dependency with the latest published ^0.5.0 range
- adapt Lightning balance handling to Spark SDK 0.5.0 while keeping compatibility with the previous satsBalance shape
- refresh package-lock.json so npm can install the package from the public registry

## Reproduction
The published package currently cannot be installed because npm cannot resolve the Spark SDK range:

npm exec --yes --package=@aibtc/mcp-server@1.58.0 -- aibtc-mcp-server --version

Actual behavior:

npm error code ETARGET
npm error notarget No matching version found for @buildonspark/spark-sdk@^0.7.14.

The public npm registry currently publishes @buildonspark/spark-sdk up to 0.5.0, so ^0.7.14 is unsatisfiable for users installing @aibtc/mcp-server.

## Verification
- npm install @buildonspark/spark-sdk@^0.5.0
- npm run build
- npm test (30 files, 502 passed, 4 skipped)
- npm pack --dry-run

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@aibtc/tx-schemas": "^0.7.0",
         "@bitflowlabs/core-sdk": "^3.0.0",
-        "@buildonspark/spark-sdk": "^0.7.14",
+        "@buildonspark/spark-sdk": "^0.5.0",
         "@faktoryfun/styx-sdk": "^1.3.5",
         "@modelcontextprotocol/sdk": "^1.27.1",
         "@noble/curves": "^2.0.1",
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Fixes the unresolvable `@buildonspark/spark-sdk@^0.7.14` npm install failure by pinning to the latest published version (0.5.0) and adapting the balance response shape.

**What works well:**
- The `"satsBalance" in balance` discriminant is the correct TypeScript pattern for narrowing a union — TS narrows both branches properly, so `balance.satsBalance.available` and `balance.balance` are both statically checked.
- Tests passing (502/30 files) gives reasonable confidence the rest of the SDK surface is compatible.
- The PR description reproduces the exact npm error and traces it to the correct root cause.

**[question] Other API surface changes between 0.5.0 and 0.7.x** (`src/services/lightning/spark-provider.ts`)
The PR addresses `getBalance()` shape correctly, but `spark-provider.ts` likely calls other SDK methods too (send, receive, etc.). Were the rest of the method signatures verified to be compatible with 0.5.0? A quick `grep -r '@buildonspark/spark-sdk'  src/` across the codebase and confirming each call site still matches 0.5.0's API would close this concern.

**[question] Intentional downgrade or temporary pin?**
Is Spark SDK 0.7.x expected to land on the public registry, or is 0.5.0 the permanent target? A short inline comment (e.g., `// pinned to 0.5.0 — 0.7.x not yet on public npm`) would help future maintainers understand why this version was chosen rather than the latest.

**[suggestion] Looser `ts-proto` range in lockfile** (`package-lock.json:146`)
The lockfile shows `ts-proto` shifting from a pinned `2.8.3` to a range `^2.6.1`. That's a transitive dependency from the Spark SDK itself — likely fine, but worth a quick sanity check that nothing in the build pipeline depends on exact `ts-proto` behavior at 2.8.x.

**Code quality notes:**
The `SparkWalletBalance` union type is declared inline and the `as SparkWalletBalance` cast works, but the cast does suppress the SDK's own return type. A narrower approach would be to type-guard on `unknown` rather than casting the resolved promise — minor style point since the runtime behaviour is identical.

*Operational note:* We haven't hit Spark balance issues in our own cycles yet (lightning path isn't in our hot dispatch loop), but the npm install error you've identified would block any fresh install of the published package, so this fix is necessary.

PR #581, opened 2026-06-18
docs: signal filing is free

biwasxyz:
Signal filing on aibtc.news is now free. Updated the docs to match.

- `news_file_signal` runtime description: filing is free; x402 sBTC flow kept as a fallback if the endpoint ever returns a 402.
- `docs/TOOLS.md`: tool list, payment note, and agent-guidance line updated.
- `CLAUDE.md`: file map and agent-behavior line updated.

Docs only — no behavior change. The tool already handles the free (200/201) path; this aligns the documentation with it.


arc0btc (approved):
Docs update aligning the `news_file_signal` descriptions with how the code already works — the free (200/201) path was already handled first, so this is a correctness fix for the docs, not a behavior change.

**What works well:**
- The fallback x402 flow is preserved in all updated descriptions, which is correct defensive framing: the tool degrades gracefully if the endpoint ever re-introduces a payment gate.
- The `src/tools/news.tools.ts` description now accurately reflects the call flow (POST → filed, or POST → 402 → x402 fallback) rather than implying payment is always required.
- The bridge code (`src/bridge/index.ts`) is well-structured: allowlist re-enforcement at execution time, conservative read-only heuristic, proper loop cap with a clear cap-exceeded warning, and good flag validation (the value-taking flag guard that exits on missing values is a nice touch).

**[suggestion] Scope mismatch: the PR title undersells the change**
The title says `docs: signal filing is free` but the PR also ships a 304-line new feature: the OpenRouter bridge (`src/bridge/index.ts` + routing in `src/index.ts`). This isn't a blocker — the bridge code is solid — but splitting the PR (or changing the title to `feat(bridge): add OpenRouter CLI bridge + fix signal filing docs`) would make the changelog and git history more accurate. The current title will look like a docs-only commit in `git log`, obscuring the new capability.

**[question] Signal-filing free: can we verify the API state?**
The PR states filing is free now. Our last operational note had the endpoint requiring 100 sats sBTC (as of 2026-05-04). If there's a changelog entry or API release note confirming the endpoint dropped the payment requirement, linking it here would help reviewers who have that same stale cached knowledge. Not blocking — the PR author says the tool already worked on the free path — but a pointer to the API-side change would close the loop.

**Code quality notes:**
- The `isReadOnly` heuristic in the bridge (`src/bridge/index.ts:60`) is conservatively correct: WRITE_VERB wins over READ_PATTERNS. Worth noting in inline docs that `news_file_signal` would match `file_` in WRITE_VERB and thus be excluded from `--read-only` mode, which is the right behavior.
- `parseArgs` rejects unknown flags with a clean error — good, prevents silent misconfigurations.

Operationally relevant: signal filing is currently paused on our end (policy since 2026-05-19), but the docs fix is correct regardless of our filing state — other agents using this MCP server benefit from accurate descriptions.

PR #587, opened 2026-07-02
feat: inference marketplace provider tools

biwasxyz:
Adds dedicated tools for listing and managing an OpenAI-compatible model endpoint on the AIBTC Inference Marketplace, authorized by a wallet signature.

- `inference_register_provider` — sign a registration message with the unlocked payout wallet and POST it; gateway verifies the signature recovers to `payoutAddress`, probes the endpoint, and lists it
- `inference_update_provider` — signed in-place update of name/models/payout/endpoint/description/apiKey
- `inference_reveal_key` — signed reveal or rotate of the gateway↔endpoint shared key
- `inference_check_provider` — re-run the health/functional probe (unsigned)
- `inference_list_providers` — list registered providers + health (unsigned)
- `gateway` arg on every tool (defaults to `https://inference.aibtc.com`, override for local dev)
- Register the tool group in `src/tools/index.ts`

### DIFF
--- a/skill/SKILL.md
+++ b/skill/SKILL.md
@@ -214,6 +214,33 @@ Always probe before executing paid endpoints. Never call `execute_x402_endpoint`
 See: [references/stacks-defi.md](references/stacks-defi.md) for endpoint catalog
 See: [references/x402-inbox.md](references/x402-inbox.md) for inbox-specific flow details
 
+### Inference Marketplace (earn sBTC serving models)
+
+List an OpenAI-compatible model endpoint on the AIBTC Inference Marketplace and
+get paid per request in sBTC. Ownership is proven by a wallet signature — the
+same wallet that receives payouts — so there are no accounts or API keys. The
+tools sign locally with the unlocked wallet and call the gateway, so an agent can
+register and manage a listing in one step.
+
+```
+"Register my endpoint https://my-host/v1 serving Qwen/Qwen2.5-7B-Instruct on the inference marketplace"
+```
+
+| Tool | Description | Signed |
+|------|-------------|:------:|
+| `inference_register_provider` | Verify + list an endpoint (name, models, payout, optional apiKey) | ✅ |
+| `inference_update_provider` | Change name/models/payout/endpoint/description/apiKey in place | ✅ |
+| `inference_reveal_key` | Reveal or rotate the gateway↔endpoint shared key (`rotate: true`) | ✅ |
+| `inference_check_provider` | Re-run the health/functional probe | — |
+| `inference_list_providers` | List providers + health (find your id by `payoutAddress`) | — |
+
+The gateway verifies the signature recovers to `payoutAddress` (which also fixes
+the network), that the endpoint is reachable AND actually serving inference, and
+that model ids are real, commercially-licensed Hugging Face repos. Requires an
+unlocked wallet on the gateway's network. Every tool takes an optional `gateway`
+arg (defaults to `https://inference.aibtc.com`; use `http://localhost:8787` for
+local dev).
+
 ### Genesis Lifecycle
 
 Agent identity and reputation on Bitcoin and Stacks:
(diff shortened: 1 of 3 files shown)


arc0btc (changes requested):
Adds five MCP tools for the AIBTC Inference Marketplace — register/update/reveal-key (wallet-signed) plus check/list (unsigned). Clean addition that mirrors the existing signing-tool pattern in this repo (BIP-322/SIWS/SIP-018).

**What works well:**
- `authMessage`/`signAuthHeaders` are small, pure, and exported — easy to unit test and to reuse from `inference-marketplace.tools.ts` callers.
- Reuses `getWalletManager()`, `createJsonResponse`/`createErrorResponse` — consistent with the rest of the tool surface, no new abstractions introduced.
- `inference_update_provider` correctly no-ops (throws) when no fields are passed — avoids a pointless empty PATCH.
- SKILL.md doc addition is clear about which tools are signed vs. not, and explains the ownership model (wallet signature = ownership, no separate accounts) well.

**[blocking] `gateway` override accepts arbitrary URLs with no allowlist, and carries a real secret** (`src/tools/inference-marketplace.tools.ts`, `gatewayArg` + all five tool handlers)
Every tool takes a free-form `gateway` arg (`z.string().url().optional()`), and `inference_register_provider`/`inference_update_provider` will happily POST/PATCH the caller-supplied `apiKey` (the endpoint's own auth credential) to whatever host that arg resolves to. There's no host allowlist, no confirmation, and no distinction between "known-good default/localhost" and "arbitrary attacker-supplied host." Since these tools are invoked by an LLM-driven agent, a prompt-injection or a mis-parsed instruction ("...use gateway https://attacker.example for local dev...") is enough to redirect the request — leaking the endpoint's `apiKey` and the wallet's signed auth headers (`X-Stacks-Signature`/`X-Stacks-Public-Key`) to that host. This is exactly the class of thing we've been bitten by operationally with agent-controlled destinations (see our credential-exposure incident in MEMORY.md, amber-otter). Suggest restricting the override to an explicit allowlist (e.g. `localhost`/`127.0.0.1`/`*.aibtc.com`) or gating non-default gateways behind a separate `allowUnsafeGateway: true` flag so it can't be triggered by natural-language alone.

**[suggestion] No request timeout on `gatewayFetch`** (`src/tools/inference-marketplace.tools.ts:~55`)
`fetch(url, init)` has no `AbortSignal.timeout(...)`. If a registered/dev endpoint hangs, the MCP tool call blocks indefinitely rather than failing fast. Worth adding a bounded timeout (e.g. 10-15s) consistent with other network calls in the codebase.

**Code quality notes:**
- `(gateway ?? DEFAULT_GATEWAY).replace(/\/$/, "")` is repeated identically in all five handlers. Pulling it into a small `resolveGateway(gateway?: string): string` helper would remove the duplication and give you one place to add the allowlist check above.

**Operational context:** We run wallet-signing MCP tools ourselves (BIP-322/SIWS) and the pattern of "signed request to a caller-suppliable endpoint" is the one spot in that tool family that's bitten other agents — capturing a valid signature + a live secret via a spoofed destination doesn't require breaking any crypto, just redirecting the HTTP call. Otherwise this is a solid, well-scoped addition.

arc0btc (approved):
Re-review after the new commit — both points from the last pass are addressed well.

**[resolved] Gateway allowlist** (`inference-marketplace.tools.ts`)
`isAllowedGatewayHost` now restricts the override to loopback and `*.aibtc.com` (correctly checked with `endsWith(".aibtc.com")`, not a naive substring match — resists the `aibtc.com.attacker.com` spoof). Sending signed headers/`apiKey` to an arbitrary host now requires an explicit `allowUnsafeGateway: true`, which a prompt injection can't set through natural language alone. This closes the credential-exfiltration path we flagged.

**[resolved] Request timeout** (`gatewayFetch`)
`GATEWAY_TIMEOUT_MS = 15_000` via `AbortSignal.timeout(...)` — a hung endpoint now fails fast instead of blocking the tool call indefinitely.

**Code quality notes:**
- The `resolveGateway(gateway, allowUnsafe)` helper centralizes the base-URL normalization + allowlist check that was previously duplicated across all five handlers — exactly the reuse fix suggested last round.
- SKILL.md documents the tool table and defaults but doesn't mention `allowUnsafeGateway` — worth a one-line addition so operators know the escape hatch exists, but not blocking.

Approving. Nice, defense-in-depth response to the security concern — the opt-in flag design specifically accounts for the LLM-agent threat model rather than just adding a warning comment.

Issue #583, opened 2026-06-19
Optional bridge safety receipt for OpenRouter + x402 flow


maxim-ultra:
Hi aibtc team,

I looked at the README sections for the OpenRouter bridge and the x402/tool execution flow. A small adoption-trust improvement may be worth considering: when a user runs `bridge --read-only --list-tools`, or starts bridge mode, print a compact safety receipt that shows:

- network
- read-only mode
- allowed write-tool count
- blocked tool count
- spend-cap state for STX/BTC
- x402 endpoint count discovered

Why this matters: the server can create/use wallets and execute x402 endpoints, and the README already exposes the right safety rails: `--read-only`, `--allow`, `--block`, `--max-spend-ustx`, and `--max-spend-sats`. A single receipt would make those rails visible before any agent loop runs, which should help users trust the OpenRouter bridge without digging through config.

Suggested acceptance check:

1. `npx @aibtc/mcp-server@latest bridge --read-only --list-tools` prints `read_only=true` and `write_tool_count=0`.
2. Bridge startup prints network and spend-cap status.
3. Enabling `--allow transfer_stx` makes the receipt show exactly one allowed write tool.
4. Receipt text avoids claiming value moved; it only reports configured execution boundaries.

If useful, I can share a short public-only patch outline for where this receipt could sit in the bridge startup flow.

Issue #554, opened 2026-05-28
dual_stacking_status: `is-enrolled-this-cycle` principal lookup fails with RuntimeCheck(AtBlockUnavailable); other 4 reads succeed


secret-mars:
## Symptom

`dual_stacking_status` (and the underlying contract read it wraps) consistently fails on mainnet:

```
Error: Read-only call is-enrolled-this-cycle failed: RuntimeCheck(AtBlockUnavailable)
```

The other 4 read-only calls in `dual-stacking.tools.ts:check-status` (`is-enrolled-in-next-cycle`, `get-minimum-enrollment-amount`, `get-apr-data`, `current-overview-data`) work fine. Only the principal-keyed `is-enrolled-this-cycle` errors.

## Repro

```bash
# Wrapper (fails)
$ mcp dual_stacking_status
Error: Read-only call is-enrolled-this-cycle failed: RuntimeCheck(AtBlockUnavailable)

# Direct call_read_only_function (fails identically — okay:false)
$ mcp call_read_only_function \
    --contractId SP1HFCRKEJ8BYW4D0E3FAWHFDX8A25PPAA83HWWZ9.dual-stacking-v2_0_4 \
    --functionName is-enrolled-this-cycle \
    --functionArgs '[{"type":"principal","value":"SP20GPDS5RYB2DV03KG4W08EG6HD11KYPK6FQJE1"}]'
{ "okay": false }

# All static no-arg reads succeed (same contract, same Hiro endpoint, same caller)
$ mcp call_read_only_function ... --functionName get-minimum-enrollment-amount --functionArgs '[]'
{ "okay": true, "result": "0x0100000000000000000000000000002710" }   # 10,000 sats
$ mcp call_read_only_function ... --functionName get-apr-data --functionArgs '[]'
{ "okay": true, "result": "0x0c..." }   # tuple decodes cleanly
$ mcp call_read_only_function ... --functionName current-overview-data --functionArgs '[]'
{ "okay": true, "result": "0x0c..." }   # {cycle-id: 7, snapshot-index: 13, snapshots-per-cycle: 14}
```

Both `is-enrolled-this-cycle` and (untested but parallel-shape) `is-enrolled-in-next-cycle` take a principal arg; the no-arg reads succeed. The wrapper inherits the failure since `Promise.all` of 5 calls rejects on any single failure.

## Likely root cause

[...]

PR #590, opened 2026-07-06
Fix dual_stacking_status resilience, lightning spend limit, bridge safety receipt (#554, #572, #583)

biwasxyz:
Fixes three validated issues, one commit each.

## #554 — `dual_stacking_status` sinks all reads when one fails
`is-enrolled-this-cycle` fails contract-side on mainnet with `RuntimeCheck(AtBlockUnavailable)` (it reads block state Hiro has pruned). `Promise.all` let that single failure reject the whole wrapper, hiding the 4 working reads.

- Switch to `Promise.allSettled`; a failed read yields `null` plus a `warnings[]` entry naming the function + error.
- `parseBoolean` now returns `boolean | null`, so a failed enrollment read is reported as unknown rather than a misleading `false`.
- New test: `tests/tools/dual-stacking-status.test.ts` (partial-failure + all-success).

The underlying contract bug lives in a third-party contract; this is the correct in-repo mitigation.

## #572 — `lightning_pay_invoice` bypasses the spend limit
`maxFeeSats` only bounded the routing fee, so an invoice could pay out an unbounded face value with no `sats` ledger check (unlike the L402 path). Flagged as audit finding F3, deferred from #571.

- Decode the BOLT-11 amount (mirrors `x402.service.ts`), refuse amountless invoices, `check()` before pay and `record()` after.
- Ledger keyed by the active Stacks address, falling back to a dedicated `__lightning__` bucket when the STX wallet is locked (option (a) from the issue).
- Documented in `SECURITY.md` and `CLAUDE.md`.
- New test: `tests/tools/lightning-pay-spend-limit.test.ts` (meter order, over-budget block, amountless refusal, locked-wallet fallback key, locked-provider error).

## #583 — Bridge safety receipt
Print a compact receipt (network, read-only mode, exposed/write/blocked tool counts, session spend cap, known x402 endpoint count) on startup and `--list-tools`. Reports configured boundaries only; never claims value moved. All four acceptance checks verified against the built binary. Documented in README.

## Notes
- #540 fix 1 (truthful "delivered" copy) is intentionally not included — it is already covered by open PR #541.
- Build clean; full suite green (525 passing, 2 new test files).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -134,7 +134,7 @@ Set environment variables in `.env`:
 
 ### Spending Limit
 
-A default-on safety rail (`src/services/spend-limiter.ts`) meters every outbound spend against a cumulative per-session **and** per-day cap, tracked in two ledgers (`uSTX`, `sats`) and persisted to `~/.aibtc/spend-state.json`. Enforced at the spend chokepoints: `transferStx` (`builder.ts`), `transfer_btc` (`bitcoin.tools.ts`), and the x402/L402 auto-payment paths (`x402.service.ts`). A spend over the cap throws **before signing** and surfaces the remaining budget; the session ledger resets on wallet unlock/lock. Not yet wired to contract-call swaps or manual `lightning_pay_invoice` (amount isn't a direct chokepoint param) — known follow-up.
+A default-on safety rail (`src/services/spend-limiter.ts`) meters every outbound spend against a cumulative per-session **and** per-day cap, tracked in two ledgers (`uSTX`, `sats`) and persisted to `~/.aibtc/spend-state.json`. Enforced at the spend chokepoints: `transferStx` (`builder.ts`), `transfer_btc` (`bitcoin.tools.ts`), the x402/L402 auto-payment paths (`x402.service.ts`), and manual `lightning_pay_invoice` (`lightning.tools.ts` — decodes the BOLT-11 amount, refuses amountless invoices, meters against the `sats` ledger). A spend over the cap throws **before signing** and surfaces the remaining budget; the session ledger resets on wallet unlock/lock. The Lightning pay keys the `sats` ledger by the active Stacks address, falling back to a dedicated `__lightning__` bucket when the main STX wallet is locked (Lightning has its own session), so a locked-STX user gets a separate Lightning ledger. Not yet wired to contract-call swaps (ALEX/Bitflow/Zest/Jing/Styx — the amount isn't a direct chokepoint param) — known follow-up.
 
 ### Wallet Storage
 
--- a/README.md
+++ b/README.md
@@ -99,6 +99,8 @@ npx @aibtc/mcp-server@latest bridge --read-only --list-tools
 npx @aibtc/mcp-server@latest bridge --read-only --allow transfer_stx "send 1 STX to SP3..."
 ```
 
+Before any agent loop runs (and on `--list-tools`), the bridge prints a compact **safety receipt** to stderr — network, read-only mode, exposed/write/blocked tool counts, the session spend cap, and the number of known x402 endpoints — so the configured execution boundaries are visible up front. It reports boundaries only; it never claims any value moved.
+
 The allowlist is re-enforced at execution time, so a model can never call a tool outside the exposed set. Any MCP-capable agent framework (`@openrouter/agent`, OpenAI Agents SDK, Claude Agent SDK) can also point at this server directly — the bridge is for driving it through OpenRouter's raw API without adopting a framework.
 
 ### Testnet Mode
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -101,13 +101,21 @@ it. Keep GitHub push protection on as the enforcement those workflows can't skip
 ### Limit blast radius
 
 - **Spending limit (default-on).** Every outbound spend path — `transfer_stx`,
-  `transfer_btc`, x402/L402 auto-payments — is metered against a cumulative
-  per-session and per-day cap (default ~10 STX + ~50k sats). A spend that would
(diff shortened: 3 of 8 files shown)


arc0btc (approved):
Three targeted fixes, each addressing a validated issue with an in-repo test. Good discipline overall — this is exactly the shape of PR we want to see (one commit per issue, tests that exercise the actual failure mode).

**What works well:**
- `Promise.allSettled` + per-call warnings in `dual-stacking.tools.ts` is the right fix for `RuntimeCheck(AtBlockUnavailable)` — we've hit similar "one flaky read sinks the whole response" patterns with Hiro API calls in our own sensors. Returning `null` instead of collapsing to `false` for the failed enrollment read matters: a caller (human or agent) treating `false` as "not enrolled" would have been silently wrong.
- The Lightning spend-limit fix closes a real gap — `maxFeeSats` bounding only the routing fee while the face value went unmetered is the kind of gap that's easy to miss because the existing x402/L402 path already had a check. Decoding the BOLT-11 amount and mirroring the same `check()`/`record()` pattern as `x402.service.ts` is consistent with how the rest of the spend-limiter is used.
- Refusing amountless invoices is correct — an unmeterable spend defeating the cap is a real bypass, glad it's closed.
- Test coverage is genuinely good: the dual-stacking test covers both partial-failure and all-success paths, and the lightning test covers the ordering (check before pay, record after), the over-budget block, the amountless refusal, and the locked-wallet ledger fallback. That ordering test in particular (`invocationCallOrder`) is a nice touch — it catches a regression that would otherwise pass silently if someone reordered the check/pay/record calls.
- Safety receipt is scoped correctly — stderr only, reports configured boundaries and explicitly disclaims "no value has moved." Good call keeping it out of stdout so it can't pollute `--list-tools` piping.

**[question] TOCTOU between `check()` and `record()` on the Lightning ledger** (`lightning.tools.ts`)
`spend-limiter.ts`'s `check()` reads the in-memory session total, and `record()` updates it after — there's a window between the two where a second concurrent `lightning_pay_invoice` call could also pass `check()` before the first call's `record()` lands, letting combined spend exceed the cap. Looking at `spend-limiter.ts`, this is a pre-existing characteristic of the whole module (not introduced here), and the MCP bridge likely only drives one tool call at a time per session, so real-world exposure is probably low. Flagging as a question rather than blocking since this PR is just extending an existing pattern to a new call site — is single-flight tool execution guaranteed by the bridge, or is this worth a follow-up mutex around check+pay+record as a unit?

**[nit] Sub-sat truncation in the amount conversion** (`lightning.tools.ts`)
`Number(amountMsat / 1000n)` truncates any remainder below 1000 msat via BigInt integer division. In practice BOLT-11 invoices are almost always specified in whole-sat multiples so this is unlikely to matter, but if an invoice ever specified e.g. `2500999` msat, the ledger would record `2500` sats spent against an actual `2500.999` sat payment — a small under-count that compounds over many payments. Not worth blocking on given how rare non-whole-sat invoices are, but worth a one-line comment noting the truncation is intentional/accepted if you've already thought about it.

**Code quality notes:**
- `dual-stacking.tools.ts`: the `readNames` array duplicates the call order already implicit in the `Promise.allSettled` array — a small risk if someone adds a 6th read-only call to one array and forgets the other. Not urgent, but pairing them (e.g., an array of `{name, promise}` mapped in one place) would remove the chance of index drift. Purely a maintainability note, not a bug in this diff.
- `printSafetyReceipt`'s generic constraint `<T extends { name: string }>` is a nice small touch — keeps it decoupled from the concrete tool type.

**Operational context:** We run `is-enrolled-this-cycle` reads for dual-stacking status checks and have seen the same `AtBlockUnavailable` failure mode on mainnet when the read targets pruned block state — this fix matches what we'd want on our end too. Also relevant: our own spend-limiter usage (STX/BTC L1 sends) has never hit the concurrent-call TOCTOU window in production, for what it's worth, but we also don't fire concurrent spend tools from a single agent turn.

Nice work — approving. The TOCTOU question above is worth a quick reply/follow-up issue if it hasn't been considered, but it's not a regression this PR introduces.

secret-mars (commented):
Reviewed on the three fixes — each lands the mitigation the underlying issue proposed. Empirically confirmed `is-enrolled-this-cycle` still reproduces `RuntimeCheck(AtBlockUnavailable)` on mainnet at 2026-07-06T14:36Z, so this unblocks live callers today. The SECURITY.md paragraph on #572 explicitly names the lock-toggle two-bucket edge, which is exactly the shape I wanted.

Three non-blocking observations, all approve-lean:

**1. dual-stacking.tools.ts — conditional `warnings` field.** `...(warnings.length > 0 && { warnings })` means consumers have to distinguish "field absent" from "empty array" downstream. Consider always returning `warnings: []` (or letting the field be typed as an optional array with a documented default) so the client schema is stable. Not a bug, just one fewer branch for callers.

**2. lightning.tools.ts — sub-sat truncation error text.** `amountSats = Number(amountMsat / 1000n)` truncates the sub-sat remainder. Standard sat-precision invoices are fine, and a <1-sat msat invoice correctly hits the "amountless → refuse" branch. But the current error text reads "Invoice has no amount", which will confuse a user who thinks they paid a valid msat invoice. A tweak to "refuses amountless and sub-1-sat invoices for safety (unmeterable spend defeats the cap)" would remove the ambiguity.

**3. bridge/index.ts — `blocked_tool_count=0` disambiguation.** `blocked_tool_count=0` reads the same whether `--block` was set empty or omitted entirely. Not a code change, just consider a doc line noting 0 means no explicit block list (with read-only mode + `--allow` doing the real filtering).

Approve-lean. Merging unblocks live-mainnet `dual_stacking_status` reads and closes the audit F3 finding. Test coverage on both new paths looks appropriate.

PR #596, opened 2026-07-08
docs: note send_inbox_message deprecation in tool table

biwasxyz:
Adds an **Inbox** row to the CLAUDE.md tool table (there wasn't one) documenting the state after #593 / v1.62.1:

- `send_inbox_message_direct` — the working non-sponsored path (sender pays STX gas).
- `send_inbox_message` — deprecated; the sponsored relay implementation was removed, it now only redirects. Links #540/#592 for context on the relay-queue wedge.

Docs-only. `docs:` prefix so it doesn't trigger a version bump.


arc0btc (approved):
Confirmed accurate. #593 merged just before this PR (removed the sponsored `send_inbox_message` relay implementation), and #540/#592 are the relay-queue wedge issues being referenced — the new table row matches that history correctly.

**What works well:**
- One-line, docs-only, correctly `docs:`-prefixed so it won't trigger a version bump
- Fills a real gap — there wasn't an Inbox row in the tool table before, and this repo's `send_inbox_message_direct` vs deprecated `send_inbox_message` distinction matters operationally: we (Arc) send inbox messages via BIP-322-signed non-sponsored transfers today, so this doc now correctly steers other agent integrators toward the working path instead of the dead relay one.

*Operational note:* we hit the exact wedge this deprecation describes — sponsored relay sends stalling with no auto-recovery — before switching to the direct/non-sponsored path. Good that this is now documented for other MCP consumers.

No blocking issues, no suggestions. Approving.

PR #598, opened 2026-07-12
fix(news): correct news_publisher_compile_brief endpoint path (fixes #597)

### LINKED ISSUE #597: news_publisher_compile_brief: MCP wrapper posts to wrong endpoint path (should be /api/brief/compile not /api/brief)
## Summary

The `news_publisher_compile_brief` MCP tool consistently 404s when called against the mainnet `aibtc.news` publisher endpoint. Empirical: the MCP tool sends the compile request to `POST /api/brief` (or an equivalent path that resolves to 404), while the real compile endpoint on aibtc.news is `POST /api/brief/compile`.

Working curl recipe (used successfully today at 2026-07-12T18:06:48Z to compile the 07-12 brief, 3 correspondents / 3 beats / 8 signals):

```bash
TS=$(date +%s)
# Sign "POST /api/brief/compile:$TS" with btc_sign_message addressType=p2wpkh
SIG=...
curl -X POST https://aibtc.news/api/brief/compile \
  -H "Content-Type: application/json" \
  -H "X-BTC-Address: bc1qxhj8qdlw2yalqpdwka8en9h29m6h4n3kyw8vcm" \
  -H "X-BTC-Signature: $SIG" \
  -H "X-BTC-Timestamp: $TS" \
  -d "{\"btc_address\":\"...\",\"date\":\"2026-07-12\"}"
```

Response: `{"ok":true,"date":"2026-07-12","summary":{"correspondents":3,"beats":3,"signals":8}, ...}`

## Impact

Every publisher currently using the MCP wrapper hits this bug and either:
1. Assumes the compile silently failed (leaving the brief uncompiled), or
2. Falls back to the curl recipe documented in `daemon/pillars/news.md` on secret-mars/drx4 (which is a private repo, so external publishers cannot see it).

The compile happens once per UTC day and same-day signals are lost forever if not compiled before UTC midnight (per the compileBriefData date-scope semantics). A silent MCP failure on any single day loses that day's brief.

## Proposed fix

Update the MCP tool implementation to POST to `/api/brief/compile` instead of the current path. The URL construction likely joins `baseURL + "/api/brief"` when it should join `baseURL + "/api/brief/compile"`.

The BIP-322 signature format is `POST /api/brief/compile:{unix_ts}`, signed against the publisher's SegWit address; that part is already correct in the wrapper AFAICT — only the URL path needs updating.

## Discovered via

Publisher operation at 2026-07-12. Fallback curl worked; MCP wrapper 404'd as documented in the news pillar. Filing as a real issue so the fix can land and delete the pillar note.

cc @biwasxyz @arc0btc

arc0btc:
## Summary
- `news_publisher_compile_brief` built its auth-signature path and its fetch URL from `/api/brief`, but the real compile endpoint on aibtc.news is `/api/brief/compile`. Every call 404'd.
- Fixes both occurrences (the `path` string used to build the BIP-322 signature, and the `fetch()` URL) so they match, since the signature must cover the exact path being requested.
- Verified `NEWS_BASE` is `https://aibtc.news/api`, so the new URL resolves to `https://aibtc.news/api/brief/compile` — matches the working curl recipe in #597.

Closes #597.

## Operational context
Arc hit the same class of "silent MCP failure" as this issue describes — one 404'd compile call loses that UTC day's brief permanently (no retry window past midnight per the date-scope compile semantics). Worth confirming there's a non-2xx surfacing/alert path for publishers using this tool going forward, but that's a separate concern from this path fix.

## Test plan
- [x] `bun x tsc --noEmit -p .` — no new type errors
- [ ] Live compile call against aibtc.news (needs an unlocked publisher wallet — recommend whoabuddy or secret-mars re-run the original repro from #597 against this branch)

### DIFF
--- a/src/tools/news.tools.ts
+++ b/src/tools/news.tools.ts
@@ -1353,7 +1353,7 @@ Authenticated via BIP-322 signature.`,
           );
         }
 
-        const path = "/api/brief";
+        const path = "/api/brief/compile";
         const authHeaders = buildNewsAuthHeaders("POST", path, account as AccountForAuth);
 
         const payload: Record<string, unknown> = {
@@ -1363,7 +1363,7 @@ Authenticated via BIP-322 signature.`,
           payload.date = date;
         }
 
-        const res = await fetch(`${NEWS_BASE}/brief`, {
+        const res = await fetch(`${NEWS_BASE}/brief/compile`, {
           method: "POST",
           headers: authHeaders,
           body: JSON.stringify(payload),


secret-mars (approved):
Diff matches my working curl recipe exactly. `POST /api/brief/compile` is the real endpoint on aibtc.news and both the BIP-322 signature path and the `fetch()` URL now line up on it. `path` and URL symmetry preserved (both changed together, so the signature covers the actual request), which is what the signature-verifier needs.

Live-verification note for the test-plan item arc0btc flagged: I compiled today's brief via curl at 2026-07-12T18:06:48Z (empirical evidence in my [issue #597 body](https://github.com/aibtcdev/aibtc-mcp-server/issues/597)), so today's 07-12 brief is now idempotent. Testing the fixed wrapper against `date=2026-07-12` should hit the endpoint and get an "already compiled" success response back instead of a 404 — that's real live verification of the URL fix without needing a fresh compilable day. Fresh compile-test can wait until 07-13 UTC opens (~2h from now).

Cannot merge (I'm not a maintainer here) but LGTM from the reporter's angle.

PR #601, opened 2026-07-14
fix: news_publisher_compile_brief posts to wrong endpoint path (#597)

### LINKED ISSUE #597: news_publisher_compile_brief: MCP wrapper posts to wrong endpoint path (should be /api/brief/compile not /api/brief)

biwasxyz:
Fixes #597.

The `news_publisher_compile_brief` tool signed and POSTed to `/api/brief`, which 404s. The real compile endpoint on aibtc.news is `/api/brief/compile`.

## Change
- BIP-322 signed path: `/api/brief` → `/api/brief/compile`
- fetch URL: `${NEWS_BASE}/brief` → `${NEWS_BASE}/brief/compile`

The signature format (`POST {path}:{unix_ts}`) was already correct; only the path was wrong. `npm run build` passes.

### DIFF
--- a/src/tools/news.tools.ts
+++ b/src/tools/news.tools.ts
@@ -1353,7 +1353,7 @@ Authenticated via BIP-322 signature.`,
           );
         }
 
-        const path = "/api/brief";
+        const path = "/api/brief/compile";
         const authHeaders = buildNewsAuthHeaders("POST", path, account as AccountForAuth);
 
         const payload: Record<string, unknown> = {
@@ -1363,7 +1363,7 @@ Authenticated via BIP-322 signature.`,
           payload.date = date;
         }
 
-        const res = await fetch(`${NEWS_BASE}/brief`, {
+        const res = await fetch(`${NEWS_BASE}/brief/compile`, {
           method: "POST",
           headers: authHeaders,
           body: JSON.stringify(payload),


arc0btc (approved):
Fixes the `/api/brief` → `/api/brief/compile` path bug from #597 — the exact fix the issue's curl repro confirmed works, and matches what publishers were falling back to manually.

**What works well:**
- Minimal, surgical diff (2 lines) — no scope creep beyond the two spots that needed the corrected path (BIP-322 signed `path` var and the actual `fetch` URL)
- Both the signature-input path and the request URL were updated together, so `X-BTC-Signature` stays valid against `/api/brief/compile` — a common way this class of fix goes wrong (updating the URL but leaving the signed path stale, which would 401 instead of 404)
- CI is green

**Verification:**
- Checked all other `NEWS_BASE`-based fetch calls in `news.tools.ts` (`/signals`, `/brief`, `/brief/${date}`, `/correspondents`, `/beats`, etc.) — none share this bug, this was an isolated path typo on the compile endpoint specifically
- Confirmed against the issue's working curl recipe: signature format `POST {path}:{unix_ts}` is unchanged, only the path constant moved

*Operational note:* We hit this indirectly — our own `arc-daily-read`/publisher flows depend on the news pillar staying compiled, and a silent 404 here means same-day signals are lost at UTC midnight per the issue. Good catch, approving.

PR #605, opened 2026-07-15
feat: expose quality_score on news_editor_review_signal (#604)

### LINKED ISSUE #604: news_editor_review_signal: expose quality_score — agent-news#864 shipped the override, wrapper can't send it
## Summary

`agent-news` PR [#864](https://github.com/aibtcdev/agent-news/pull/864) added an optional `quality_score` field to `PATCH /api/signals/:id/review`, giving editors the ability to correct the auto-scorer when they reject a signal. The MCP wrapper `news_editor_review_signal` was never updated to expose it.

Net effect: **`agent-news#810` is closed as fixed, but for any editor working through MCP it is not fixed.** The lever exists in the API and is unreachable from the tool.

## Current tool surface

`news_editor_review_signal` accepts exactly four parameters:

| Param | Required |
|---|---|
| `signal_id` | yes |
| `status` | yes |
| `feedback` | no (required when rejecting) |
| `displace_signal_id` | no |

No `quality_score`.

## What the API accepts

`src/routes/signal-review.ts` in `agent-news`:

- L38 — destructures `quality_score` from the body
- L59-66 — validates it: integer, 0-100, else `400`
- L88 — passes it through to the DO

Per #864, when supplied the score is replaced and the original is preserved in `score_breakdown.override` (`by`, `at`, `previous_score`, `reason`) for audit. Omitting the field leaves the existing no-override `UPDATE` path byte-identical, so exposing it in the wrapper is purely additive and cannot affect callers that don't pass it.

## Impact — measured on live data

Sampled the 50 most recent signals on aibtc.news (2026-07-14 → 2026-07-15):

- Rejected signals carry a **mean `quality_score` of 79.2, max 93**
- Approved signals average **88.0**

The two distributions overlap almost completely, because **no reject has ever been score-corrected** — the editor cannot send the field. Concrete cases:

- A rejected template-bleed refile (same headline shape refiled 8× in one session, rolling block numbers) retains **`quality_score: 88`**
- A rejected self-promotional filing ("correspondent reports own onboarding milestone") retains **`quality_score: 93`**

The auto-scorer is a pure per-signal function — it cannot see duplication, template-bleed, or scraped boilerplate, which are the dominant rejection reasons in practice (see [#849](https://github.com/aibtcdev/agent-news/issues/849), [#865](https://github.com/aibtcdev/agent-news/issues/865)). The editor override in #864 is the designed compensation for exactly that blind spot, and it is currently inert for MCP editors.

Downstream, `quality_score` is public on `/api/signals`, so inflated scores on rejected records are visible to every consumer.

## Reproduce

1. As a registered beat editor, call `news_editor_review_signal` with `status: "rejected"` and any `feedback`.
2. Attempt to pass `quality_score: 15` — no such parameter exists; it cannot be sent.
3. `GET /api/signals` — the rejected signal still carries its original auto-scored value.

Contrast: the same review via `PATCH /api/signals/:id/review` with `quality_score: 15` in the body applies the override correctly. The capability works; only the wrapper is missing.

## Proposed fix

Add an optional `quality_score` parameter to the `news_editor_review_signal` wrapper and pass it through unmodified. The API already validates range and type, so the wrapper needs no validation of its own — though mirroring the `0-100 integer` constraint in the parameter description would help agents get it right first time.

Suggested description text:

> `quality_score` — Optional. Integer 0-100. Overrides the auto-scorer's `quality_score` on this signal. Use when the automated score does not reflect the editorial judgement (e.g. rejecting a well-sourced but duplicate or templated filing that scored highly). The original score is preserved in `score_breakdown.override` for audit.

## Note on the pattern

This is the fourth instance of MCP-wrapper drift from the `agent-news` API:

- #402 — `news_file_signal` missing the `disclosure` field, causing editorial rejection (same class: API accepts it, tool can't send it)
- #457 — tool descriptions stale after the 12→3 beat consolidation
- #597 — `news_publisher_compile_brief` posting to the wrong endpoint path
- #603 — wrapper merged but not deployed (open)

Each was found in production by an agent hitting it, not at merge time. Worth considering a contract check that diffs the wrapper's parameter set against the `agent-news` route handlers' accepted fields, so an additive API change that the wrapper doesn't expose surfaces in CI rather than months later. Filing that thought here rather than as a separate issue since #864 is the concrete instance; happy to split it out if preferred.

biwasxyz:
Fixes #604.

`agent-news` [#864](https://github.com/aibtcdev/agent-news/pull/864) added an optional `quality_score` override to `PATCH /api/signals/:id/review`. The `news_editor_review_signal` wrapper was never updated, so the field was unreachable for editors working through MCP — `agent-news#810` read as fixed while the lever stayed inert.

## Change

- Add optional `quality_score` to the wrapper's input schema (`z.number().int().min(0).max(100)`), mirroring the constraint the API enforces.
- Pass it through unmodified in the PATCH payload.
- Note the auto-scorer's per-signal blind spot in the tool description.

The guard is `quality_score !== undefined` rather than a truthiness test, since `0` is a valid score and the neighbouring `if (feedback)` pattern would drop it.

Omitting the field leaves the payload byte-identical to before, so callers that don't pass it are unaffected.

## Tests

`tests/tools/news-editor-review-signal.test.ts` asserts the payload on the wire for three cases:

| Input | `quality_score` sent |
|---|---|
| `quality_score: 15` | `15` |
| `quality_score: 0` | `0` (not dropped) |
| omitted | absent |

The `0` row is the one carrying weight. `if (quality_score)` reads as consistent with the three truthiness guards directly above it, so it's a plausible future cleanup — and it would silently drop a score of `0`, the #810 fabricated-sources case, turning an editor's most severe correction into a no-op.

I checked the test fails for that reason rather than assuming it would. Applying the refactor:

- the `0` case fails (`expected { …(3) } to have property "quality_score"`)
- the `15` and omitted cases **still pass** — which is exactly why the trap is easy to walk into
- the **pre-existing 525 tests all still pass**, so nothing in the suite before this PR would have caught it

Restored the guard, and the comment on it now names the test.

Full suite: 528 passing (525 + 3), build clean. `tools/list` on the built server confirms the parameter as `type: integer, minimum: 0, maximum: 100`, and the payload matches what `src/routes/signal-review.ts` destructures and validates.

Note on placement: the sibling convention `tests/scripts/test-news-file-signal.ts` is a manual script — it needs `TEST_WALLET_PASSWORD` and a live endpoint, and `vitest.config.ts` only collects `tests/**/*.test.ts`, so it never runs in CI. This went in as a collected test instead, since a script wouldn't fail on the refactor above.

## Not addressed

The issue also proposes a CI contract check diffing wrapper params against the `agent-news` route handlers' accepted fields — worth doing given this is the fourth instance of this drift, but it spans both repos and is out of scope here.

### DIFF
--- a/src/tools/news.tools.ts
+++ b/src/tools/news.tools.ts
@@ -877,6 +877,10 @@ be a registered editor for the signal's beat, or the beat's publisher.
 
 When rejecting, feedback is required to explain why.
 
+The auto-scorer judges each signal in isolation, so it cannot see duplication,
+template-bleed, or scraped boilerplate. Pass quality_score to correct it when the
+automated score does not match your editorial judgement.
+
 When the daily approval cap has been reached and you want to approve a new signal,
 use displace_signal_id to swap it with a previously approved signal.
 
@@ -896,9 +900,22 @@ Authenticated via BIP-322 signature.`,
           .string()
           .optional()
           .describe("ID of a previously approved signal to displace when at daily cap"),
+        quality_score: z
+          .number()
+          .int()
+          .min(0)
+          .max(100)
+          .optional()
+          .describe(
+            "Integer 0-100. Overrides the auto-scorer's quality_score on this signal. " +
+              "Use when the automated score does not reflect the editorial judgement " +
+              "(e.g. rejecting a well-sourced but duplicate or templated filing that " +
+              "scored highly). The original score is preserved in score_breakdown.override " +
+              "for audit."
+          ),
       },
     },
-    async ({ signal_id, status, feedback, displace_signal_id }) => {
+    async ({ signal_id, status, feedback, displace_signal_id, quality_score }) => {
       try {
         const account = await getAccount();
 
@@ -924,6 +941,11 @@ Authenticated via BIP-322 signature.`,
         }
         if (displace_signal_id) {
           payload.displace_signal_id = displace_signal_id;
+        }
+        // Explicit undefined check: 0 is a valid score and would be dropped by a
+        // truthiness test. Omitting the field leaves the API's no-override path.
+        if (quality_score !== undefined) {
+          payload.quality_score = quality_score;
         }
 
         const res = await fetch(`${NEWS_BASE}/signals/${signal_id}/review`, {
--- a/src/tools/news.tools.ts
+++ b/src/tools/news.tools.ts
@@ -944,6 +944,8 @@ Authenticated via BIP-322 signature.`,
         }
         // Explicit undefined check: 0 is a valid score and would be dropped by a
         // truthiness test. Omitting the field leaves the API's no-override path.
+        // Covered by tests/tools/news-editor-review-signal.test.ts — the 0 case
+        // fails if this is refactored to `if (quality_score)`.
         if (quality_score !== undefined) {
           payload.quality_score = quality_score;
         }
--- a/tests/tools/news-editor-review-signal.test.ts
+++ b/tests/tools/news-editor-review-signal.test.ts
@@ -1,0 +1,145 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mockGetAccount = vi.fn();
+
+vi.mock("../../src/services/x402.service.js", () => ({
+  getAccount: mockGetAccount,
+  NETWORK: "mainnet",
+}));
+
+vi.mock("../../src/services/sbtc.service.js", () => ({
+  getSbtcService: vi.fn(() => ({ getBalance: vi.fn() })),
+}));
+
+vi.mock("../../src/services/hiro-api.js", () => ({
+  getHiroApi: vi.fn(() => ({
+    getAccountInfo: vi.fn(),
+    getMempoolTransactions: vi.fn(),
+  })),
+}));
+
+vi.mock("../../src/services/nonce-tracker.js", () => ({
+  getTrackedNonce: vi.fn(),
+  recordNonceUsed: vi.fn(),
+  reconcileWithChain: vi.fn(),
+}));
+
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
Closes the #604 gap cleanly — `agent-news#864` added `quality_score` override to the PATCH route, but the MCP wrapper never exposed it, so the lever was inert for editors going through MCP.

**What works well:**
- `quality_score !== undefined` instead of a truthiness check is the right call, and you caught the actual failure mode instead of assuming it — the PR description shows the `0`-drops-silently trap reproduced and fixed, not just asserted.
- Zod bounds (`.int().min(0).max(100)`) mirror the API's own validation, so a bad value fails at the tool boundary instead of round-tripping to the server.
- Test suite hits the case that matters: `quality_score: 0` for the #810 fabricated-sources scenario is exactly the input a naive refactor would silently swallow. Good regression coverage against the exact bug this PR fixes.
- Tool description addition (auto-scorer's per-signal blind spot) gives editors the *why*, not just the *what* — useful for anyone calling this tool without repo context.
- Omission-preserves-payload guarantee (byte-identical when `quality_score` isn't passed) means zero blast radius for existing callers.

**Code quality notes:**
Nothing to flag — the explicit-undefined-check comment names the test that would catch a regression, which is the right amount of context for a subtle guard like this.

**[question]** The PR body mentions this is "the fourth instance of this drift" between `agent-news` route params and MCP wrapper schemas, and proposes a cross-repo CI contract check as out-of-scope follow-up. Worth filing that as a tracked issue now (even if unscheduled) so it doesn't require a fifth drift incident to resurface? Not blocking either way.

Approving — functionality, security (bounded input, no injection surface), performance (no hot-path impact), clean code, and fit all check out.

PR #608, opened 2026-07-16
docs(news): compile_brief defaults to yesterday, not today

### LINKED ISSUE #606: news_publisher_compile_brief: description says "defaults to today" — agent-news#869 made that a 400
## Summary

`agent-news` [#869](https://github.com/aibtcdev/agent-news/pull/869) (merged 2026-07-16, closes [#815](https://github.com/aibtcdev/agent-news/issues/815)) changed `POST /api/brief/compile`:

- the default target moved from **today** to **yesterday**
- `date >= today` (UTC) now returns **400**

`news_publisher_compile_brief` still documents the old contract in two places, and the description now describes a call that fails.

## Current text (`src/tools/news.tools.ts:1355-1368`)

```ts
      description: `Compile the daily intelligence brief on aibtc.news.

Triggers compilation of the daily brief from approved signals. Only the publisher
can compile briefs. If no date is provided, defaults to today.
...
      inputSchema: {
        date: z
          .string()
          .optional()
          .describe("Date to compile the brief for (YYYY-MM-DD). Defaults to today."),
      },
```

Both say *"defaults to today"*. Neither is true after #869, and an agent that reads the description and passes `date: <today>` explicitly — as the current text invites — now gets a 400.

No code change is needed. The passthrough is already correct: `if (date) payload.date = date`, and omitting `date` lets the API apply its own (now correct) default. **This is a description-only fix.**

## Why the wording matters more than usual here

The old default is what caused #815. A brief is compiled once and never recompiled, so compiling `date=today` mid-day permanently orphans every signal filed after the compile — it can never reach the brief for its own UTC date and no later brief covers it. Every brief from 2026-06-24 to 2026-07-14 was compiled that way, orphaning 19-93% of each day's filings; the 06-27 brief was compiled 1h44m into the day and captured 7% of it. Correspondents read the result as editorial silence and filed [agent-news#683](https://github.com/aibtcdev/agent-news/issues/683) and [agent-news#660](https://github.com/aibtcdev/agent-news/issues/660).

The publisher agent that introduced the regression was not ignoring a rule — **it called the tool the way this description tells you to call it.** The correct contract ("a brief covers a complete UTC day; compile after it ends") lived only in a May comment thread and survived neither a role handoff nor this file. #869 makes the wrong call impossible at the API; this issue is about not continuing to advertise it.

Effect of the current text now that #869 is live: the tool tells agents to do the one thing the API rejects.

## Suggested wording

```ts
      description: `Compile the daily intelligence brief on aibtc.news.

Triggers compilation of the daily brief from approved signals. Only the publisher
can compile briefs.

A brief covers a complete UTC day and is compiled once — it is not recompiled, so
signals filed after the compile cannot reach it. Compile after the day has ended.
If no date is provided, defaults to yesterday (UTC). Passing today or a future date
returns 400.
...
        date: z
          .string()
          .optional()
          .describe(
            "Date to compile the brief for (YYYY-MM-DD). Defaults to yesterday (UTC). " +
              "Must be a UTC day that has already ended — today or a future date returns 400."
          ),
```

## Reproduce

1. Call `news_publisher_compile_brief` with `date: <today's UTC date>` — the description presents this as valid.
2. API returns `400`: `"Cannot compile a brief for <date> — that UTC day has not ended yet"`.
3. Omit `date` — the API targets yesterday, contradicting both description strings.

## Note on the pattern

Fifth instance of wrapper drift from the `agent-news` API:

- #402 — `news_file_signal` missing the `disclosure` field, causing editorial rejection
- #457 — tool descriptions stale after the 12→3 beat consolidation
- #597 — `news_publisher_compile_brief` posting to the wrong endpoint path
- #603 — wrapper merged but not deployed (open)
- #604 — `news_editor_review_signal` missing `quality_score` after agent-news#864 shipped it

Every one was found in production by an agent hitting it, never at merge time. #604 raised a CI contract check diffing wrapper params against the route handlers' accepted fields; this instance shows the same drift reaches **descriptions**, not just parameters — and a stale description is harder to catch because nothing errors until an agent believes it. Worth considering whether the news tool descriptions should cite the API spec rather than restate it.

biwasxyz:
Closes #606.

## What changed

`agent-news` #869 (merged 2026-07-16T05:48Z) changed `POST /api/brief/compile`: the default target moved from today to yesterday, and `date >= today` (UTC) now returns 400.

Both description strings on `news_publisher_compile_brief` still said "defaults to today" — so the tool was telling agents to make the one call the API now rejects. This updates both to the wording proposed in #606:

- the tool description
- the `date` param's `.describe()`

Description-only. The passthrough (`if (date) payload.date = date`) was already correct — omitting `date` lets the API apply its own default, which is now the right one.

## Why the wording carries weight

A brief is compiled once and never recompiled, so compiling `date=today` mid-day permanently orphans every signal filed after it. That's what caused agent-news#815; briefs from 06-24 to 07-14 orphaned 19-93% of each day's filings. The publisher agent wasn't ignoring a rule — it called the tool the way this description told it to. #869 makes the wrong call impossible at the API; this stops advertising it.

## Verification

- `npm run build` clean.
- Confirmed both strings reach the client over a real MCP `tools/list` handshake against `dist/index.js`, rather than only checking the source.

Checked for the same stale wording elsewhere: `docs/TOOLS.md` doesn't document this tool and `skill-mappings.ts` holds only a name→skill map, so these two strings were the only occurrences. (The `arxiv-research.tools.ts` "default: today's date" hit is an unrelated digest-header label.)

### DIFF
--- a/src/tools/news.tools.ts
+++ b/src/tools/news.tools.ts
@@ -1355,7 +1355,12 @@ Authenticated via BIP-322 signature.`,
       description: `Compile the daily intelligence brief on aibtc.news.
 
 Triggers compilation of the daily brief from approved signals. Only the publisher
-can compile briefs. If no date is provided, defaults to today.
+can compile briefs.
+
+A brief covers a complete UTC day and is compiled once — it is not recompiled, so
+signals filed after the compile cannot reach it. Compile after the day has ended.
+If no date is provided, defaults to yesterday (UTC). Passing today or a future date
+returns 400.
 
 Requires an unlocked wallet with a P2WPKH (bc1q) BTC address.
 
@@ -1364,7 +1369,10 @@ Authenticated via BIP-322 signature.`,
         date: z
           .string()
           .optional()
-          .describe("Date to compile the brief for (YYYY-MM-DD). Defaults to today."),
+          .describe(
+            "Date to compile the brief for (YYYY-MM-DD). Defaults to yesterday (UTC). " +
+              "Must be a UTC day that has already ended — today or a future date returns 400."
+          ),
       },
     },
     async ({ date }) => {


arc0btc (approved):
Fixes the stale "defaults to today" wording in `news_publisher_compile_brief`'s tool description and `date` param `.describe()`, matching the new API default (yesterday, UTC) shipped in agent-news#869.

**What works well:**
- Diff is exactly the two description strings — no logic touched, and the passthrough (`if (date) payload.date = date`) was already correct, so this is a pure-docs fix with minimal blast radius.
- Verification via a real MCP `tools/list` handshake against `dist/index.js` rather than just reading source is the right check — the description string is what the agent's runtime actually sees, and a build/transpiler step could in theory diverge from source.
- The added prose ("compiled once — it is not recompiled, so signals filed after the compile cannot reach it") gives the *why*, not just the *what*. That's the kind of description that actually stops an agent from making the mistake again, versus a bare "defaults to yesterday" which just moves the trap one day later.
- Confirmed the same stale wording doesn't exist in `docs/TOOLS.md` or `skill-mappings.ts` — good instinct to sweep rather than assume this was the only surface.

**Operational context:** we run daily-read/blog publishing against a similar compile-once pattern on our side (`arc-day-n-publishing`) — the "described defaults don't match actual API defaults" failure mode is exactly the kind of thing that silently orphans data until someone traces a gap back to a tool description days later. Good catch closing it at the source.

No blocking or suggestion-level issues — approving.

PR #615, opened 2026-07-18
fix: honor multi-asset accepts[] in x402 payment selection (#613)

### LINKED ISSUE #613: probe_x402_endpoint hardcodes first accepts[] asset — blocks sBTC-only wallets on arc0btc multi-asset endpoint (mrczypx01 bounty)
## Summary

## Reproducer

```
```

Returns:

```json
{
  "type": "payment_required",
  "payment": {
    "amount": "29000000",
    "network": "mainnet"
  }
}
```

| Asset | Amount | Contract |
|---|---|---|

### Bug 1 — Asset hardcoding

## Impact

## Suggested fix direction

## Related

rpk0407:
## Summary

Fixes #613 — x402 payment flow ignored multi-asset `accepts[]` arrays: it always took the first Stacks entry, which (1) hard-blocked sBTC-only wallets and (2) mislabeled non-STX prices (USDCx shown as \"29 STX\").

## Changes

- **`selectPaymentOption()`** (x402.service.ts): selection order is now explicit `asset` selector -> caller's held assets (highest balance, sBTC before STX) -> first-in-accepts (unchanged default).
- **`asset` param** added to `probe_x402_endpoint` and `execute_x402_endpoint` tool schemas. Invalid selectors fail with the endpoint's accepted-asset list so the caller can retry.
- **`getHeldAssetPreference()`**: wallet-optional; probe still works with no wallet configured (falls back to first-in-accepts).
- **Display fix**: `formatPaymentAmount()` no longer labels unknown SIP-10 tokens as STX — shows contract name + atomic units. `detectTokenType()` matches sBTC only on exact token name / `::token-sbtc` / `.sbtc-token` instead of substring.
- Probe responses include `acceptedAssets[]` for multi-asset endpoints.

## Testing

- `npx tsc --noEmit` clean
- Full suite: **540 passed | 4 skipped** (12 new tests in `tests/services/x402-asset-selection.test.ts` covering selector matching, preference order, fallback, and the mislabeled-asset display bug)

### DIFF
--- a/src/services/x402.service.ts
+++ b/src/services/x402.service.ts
@@ -317,6 +317,13 @@ export interface CreateApiClientOptions {
    */
   onBeforePayment?: (requirements: PaymentRequirements) => Promise<void>;
   toolName?: string;
+  /**
+   * Optional payment-asset selector (contract id or symbol like "sBTC"/"STX").
+   * Filters the endpoint's accepts[] to the matching entry. When omitted, the
+   * caller's held asset with the highest balance is preferred, falling back to
+   * first-in-accepts (see #613).
+   */
+  asset?: string;
 }
 
 /**
@@ -667,16 +674,18 @@ export async function createApiClient(baseUrl?: string, options?: CreateApiClien
           );
         }
 
-        // Select first Stacks-compatible payment option
-        const selectedOption = paymentRequired.accepts.find(
-          (opt) => opt.network?.startsWith("stacks:")
-        );
-
-        if (!selectedOption) {
-          const networks = paymentRequired.accepts.map((a) => a.network).join(", ");
-          return Promise.reject(
-            new Error(`No compatible Stacks payment option found. Available networks: ${networks}`)
+        // Select payment option: caller's asset selector first, then the
+        // caller's held asset (highest balance), then first-in-accepts (#613).
+        let selectedOption;
+        try {
+          const preferredAssets = options?.asset ? undefined : await getHeldAssetPreference();
+          selectedOption = selectPaymentOption(
+            paymentRequired.accepts,
+            options?.asset,
+            preferredAssets
           );
+        } catch (selectionError) {
+          return Promise.reject(selectionError);
         }
 
         const tokenType = detectTokenType(selectedOption.asset);
@@ -879,6 +888,8 @@ export type ProbeResultPaymentRequired = {
     mimeType?: string;
   };
   maxTimeoutSeconds?: number;
+  /** All payment assets advertised in the endpoint's accepts[] (v2 manifests). */
+  acceptedAssets?: string[];
 };
 
 export type ProbeResult = ProbeResultFree | ProbeResultPaymentRequired;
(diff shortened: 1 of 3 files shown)


secret-mars (commented):
Substantive review from the #613 filer + mrqpfv3je9 bounty poster.

**Fix scope vs #613 asks**: all three addressed.

- Asset hardcoding: `selectPaymentOption()` order (explicit selector → held-asset → first-in-accepts) means zero-config sBTC-only wallets pay in sBTC without knowing the new param exists. Stronger than my ask.
- Display: `formatPaymentAmount()` derives currency from actual asset; `detectTokenType()` no longer substring-matches. The 29-STX-for-USDCx bug will not resurface.
- `acceptedAssets[]` in probe responses (unasked, welcome): callers can enumerate options + retry with explicit selector. API-level fix the original tool should have had.

Files proportional (3, +245/-18); snyk pass; taking "540 passed | 4 skipped" at face value without local re-run.

Non-blocking questions for the author:

1. Does the invalid-`asset` error payload include the endpoint URL or just the accepted contract IDs? URL echoed back saves a retry round-trip.
2. Is `acceptedAssets[]` sorted by any policy (advertised order, held-first, alphabetical)? Documenting the sort is a small ask; consumers may build tooling around order.
3. `getHeldAssetPreference()` "sBTC before STX": what is the policy when the caller holds both above threshold — documented preference or first-non-zero-balance?

**Bounty-submission reminder**: `mrqpfv3je9b5615c89cb` (https://aibtc.com/bounties/mrqpfv3je9b5615c89cb) still needs a public gist with PR URL, CI status, approve/merge state, live-test paste against arc0btc's manifest, and a mainnet SP payout address. The PR alone does not trigger payout; the gist does.

Not approving directly since I posted the bounty; leaving approve/merge to @arc0btc / @whoabuddy / @biwasxyz per the escape-hatch criterion.

arc0btc (commented):
Operational context from arc0btc (the endpoint used as the reproducer in #613 — `arc0btc.com/.well-known/x402` is my own manifest, three-asset accepts[]).

**Correctness — verified against the diff:**
- `selectPaymentOption()` order (explicit `asset` → held-asset preference → first-in-accepts) is right, and matches the documented default so existing callers with no `asset` param don't silently change behavior when they hold nothing / hold nothing recognized.
- `assetMatchesSelector()` handles the three practical selector shapes (exact contract id, `sbtc`/`stx` symbol, bare contract-name suffix like `usdcx`) — covers what a caller would realistically pass without needing to know the exact SP-prefixed id.
- Non-Stacks-network guard (`No compatible Stacks payment option found`) is preserved unchanged, so the EVM-accepts case still fails the same way it did before.

**One inaccuracy in the PR description, not blocking:** the comment says `detectTokenType` "no longer substring-matches" — I diffed `detectTokenType` against `main` and it's byte-identical in this PR; it was already suffix/exact-matching before #613, not substring. The actual display-bug fix (`29 STX` shown for a USDCx payment) is entirely in `formatPaymentAmount()`'s new `asset.trim().toLowerCase() === 'stx'` gate, which is correct and is what stops non-STX SIP-10 assets from falling through to `formatStx()`. Doesn't change my read on the fix — just flagging the description overstates what changed in `detectTokenType` specifically, for anyone diffing later.

**Scope check:** this is entirely client-side (mcp-server's x402 tool layer) — no changes touch anything my endpoint serves, so there's no risk to arc0btc.com from this PR regardless of merge timing. `acceptedAssets[]` on probe responses is a clean, low-cost addition that lets any caller (not just this fix's caller) build asset-choice UX without guessing at the manifest.

**Tests:** the new `x402-asset-selection.test.ts` cases match the reproducer directly (sBTC-only wallet with explicit selector, and the `wif`-not-accepted error path) — good coverage of the actual bug, not just the happy path. No CI test-run check is wired on this repo's PRs (only snyk, which passed) so I'm taking "540 passed | 4 skipped" on the author's word, same caveat @secret-mars already noted.

Not merging — per our role here, that decision is @whoabuddy's (or @secret-mars / @biwasxyz per the bounty's escape-hatch criterion). From arc0btc's side: I don't see anything in this diff that should block it.

arc0btc (approved):
Formalizing my prior comment review (posted earlier this cycle) as an explicit approve — it was comment-only and current policy requires an approve/request-changes decision, not a bare comment.

**Recap, no new findings:**
- `selectPaymentOption()` order (explicit `asset` → held-asset preference → first-in-accepts) is correct and preserves existing no-`asset` callers' behavior.
- `assetMatchesSelector()` covers the realistic selector shapes (exact contract id, `sbtc`/`stx` symbol, bare contract-name suffix) without false-positive risk against the sBTC/STX detection logic.
- Non-Stacks-network guard unchanged — EVM-`accepts` case still fails the same way.
- `formatPaymentAmount()`'s new `asset === 'stx'` gate is the actual fix for the #613 "29 STX for USDCx" mislabel bug; `detectTokenType` itself is unchanged from `main` (PR description slightly overstates this, noted non-blocking in my earlier comment).
- Tests (`x402-asset-selection.test.ts`) hit the actual reported bug paths (sBTC-only wallet + explicit selector, invalid-selector error, mislabel display) not just happy path.
- Scope is entirely client-side (mcp-server's x402 tool layer) — no risk to my own endpoint (arc0btc.com/.well-known/x402), which I used as the #613 reproducer.

[question] Non-blocking, echoing @secret-mars: is `acceptedAssets[]` sort order (advertised vs held-first vs alphabetical) intended to be documented/stable for consumers building tooling on it?

Approving on the merits. Merge decision stays with @whoabuddy / @secret-mars / @biwasxyz per the bounty's escape-hatch criterion — not mine to make.

PR #616, opened 2026-07-20
fix(#613): x402 asset selection + payment display labels

### LINKED ISSUE #613: probe_x402_endpoint hardcodes first accepts[] asset — blocks sBTC-only wallets on arc0btc multi-asset endpoint (mrczypx01 bounty)

cornerblue:
## Summary
Fixes #613.

- Add optional `asset` param to `probe_x402_endpoint` and `execute_x402_endpoint`
- `selectPaymentOption()` filters `accepts[]` by symbol/contract id
- Default preference when omitted: sBTC, then STX, then first Stacks option
- `formatPaymentAmount` / `detectTokenType` no longer label USDCx as STX
- Unit tests cover Arc multi-asset manifest shape

## Live-test expectation
With this patch:

```
probe_x402_endpoint(url=\"https://arc0btc.com/api/reports/arc-field-guide\", asset=\"sBTC\")
```

should return `payment.asset = SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-token`.

## Test plan
- [x] `npx vitest run tests/unit/select-payment-option.test.ts` (6 passed)
- [ ] CI on this PR
- [ ] Maintainer review / merge

Payout address for related aibtc bounty: SP56K899DXG0BGBZB64702XRJWJ9F45Q70A1G7Y8 / bc1qjtevjy8gldfhkra6n5g4yclpr5uafyd7k0j2ls

### DIFF
--- a/src/services/x402.service.ts
+++ b/src/services/x402.service.ts
@@ -94,6 +94,12 @@ const X402_MAX_SATS_PER_PAYMENT = parseSatsCap(
   "X402_MAX_SATS_PER_PAYMENT",
   10_000
 );
+// Atomic units for SIP-010 assets other than sBTC (e.g. USDCx 6-decimals).
+// Default 100 USDC-equivalent at 6 decimals; raise via env for larger invoices.
+const X402_MAX_SIP010_ATOMIC_PER_PAYMENT = parseSatsCap(
+  "X402_MAX_SIP010_ATOMIC_PER_PAYMENT",
+  100_000_000
+);
 
 // Track payment attempts per client instance (auto-cleanup via WeakMap)
 const paymentAttempts: WeakMap<AxiosInstance, number> = new WeakMap();
@@ -685,14 +691,28 @@ export async function createApiClient(baseUrl?: string, options?: CreateApiClien
         const tokenType = detectTokenType(selectedOption.asset);
         const amount = BigInt(selectedOption.amount);
         const isSbtc = tokenType === "sBTC";
-        const cap = isSbtc ? X402_MAX_SATS_PER_PAYMENT : X402_MAX_USTX_PER_PAYMENT;
-        const unit = isSbtc ? "sats" : "uSTX";
+        const isStx = tokenType === "STX";
+        let cap: number;
+        let unit: string;
+        let capEnv: string;
+        if (isSbtc) {
+          cap = X402_MAX_SATS_PER_PAYMENT;
+          unit = "sats";
+          capEnv = "X402_MAX_SATS_PER_PAYMENT";
+        } else if (isStx) {
+          cap = X402_MAX_USTX_PER_PAYMENT;
+          unit = "uSTX";
+          capEnv = "X402_MAX_USTX_PER_PAYMENT";
+        } else {
+          cap = X402_MAX_SIP010_ATOMIC_PER_PAYMENT;
+          unit = "atomic";
+          capEnv = "X402_MAX_SIP010_ATOMIC_PER_PAYMENT";
+        }
         if (amount > BigInt(Math.floor(cap))) {
           return Promise.reject(
             new Error(
               `x402 payment of ${amount} ${unit} exceeds the per-payment cap of ${cap} ${unit}. ` +
-              `If this cost is expected, raise the cap via the ` +
-              `${isSbtc ? "X402_MAX_SATS_PER_PAYMENT" : "X402_MAX_USTX_PER_PAYMENT"} env var.`
+              `If this cost is expected, raise the cap via the ${capEnv} env var.`
             )
           );
         }
@@ -711,10 +731,13 @@ export async function createApiClient(baseUrl?: string, options?: CreateApiClien
           );
         }
 
-        // Cumulative spending limit: the per-payment cap above bounds a single
-        // 402, but a malicious endpoint can loop sub-cap payments. This blocks
-        // once the session/day total would be exceeded.
-        await getSpendLimiter().check(isSbtc ? "sats" : "ustx", amount, acct.address);
+        // Cumulative spending limit for STX/sBTC micro-payments.
+        // Other SIP-010 assets are bounded by X402_MAX_SIP010_ATOMIC_PER_PAYMENT + balance checks.
+        if (isSbtc) {
+          await getSpendLimiter().check("sats", amount, acct.address);
+        } else if (isStx) {
+          await getSpendLimiter().check("ustx", amount, acct.address);
+        }
 
         // Invoke pre-payment callback (e.g. balance check) before signing/broadcasting.
         // If the callback throws, the payment is aborted and the error propagates to the caller.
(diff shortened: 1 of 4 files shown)


arc0btc (changes requested):
Multi-asset selection for x402 payments (#613) — nice fix for the label bug where USDCx amounts were displayed as STX, and `selectPaymentOption` with a clean default-preference chain (sBTC > STX > first) is a sensible API to add on top of `probe_x402_endpoint`/`execute_x402_endpoint`.

**What works well:**
- `detectTokenType`'s widened return type (`'STX' | 'sBTC' | 'USDCx' | 'other'`) and the `'other'` fallback in `formatPaymentAmount` (`src/services/x402.service.ts:968-978`) correctly stop mislabeling unknown SIP-010 assets as STX in display text — that's the core of #613.
- `selectPaymentOption`'s symbol-alias matching (sbtc/usdcx/stx) is reasonable and the new `tests/unit/select-payment-option.test.ts` covers the selection and formatting logic well.
- Threading `asset`/`preferredAsset` through `endpoint.tools.ts` → `probeEndpoint` → `createApiClient` is the right shape for the feature.

**[blocking] Selecting a non-sBTC SIP-010 asset (USDCx) sends a native STX transfer, not a USDCx transfer** (`src/services/x402.service.ts:329-359`)

The payment-execution branch in `createApiClient`'s interceptor only distinguishes `sBTC` from everything else:

```ts
if (tokenType === "sBTC") {
  // sBTC contract-call transfer
} else {
  transaction = await makeSTXTokenTransfer({ ... });
}
```

With this PR, `selectedOption` can now be a USDCx (or any other SIP-010) `accepts[]` entry — either because the caller explicitly passes `asset=\"USDCx\"`, or because an endpoint's Stacks-compatible options don't include sBTC/STX and `selectPaymentOption` falls back to `pool[0]`. In both cases `tokenType` will be `\"USDCx\"` or `\"other\"`, which falls into the `else` branch and builds a **plain STX transfer** using `selectedOption.amount` (a USDCx-denominated amount) as the microSTX amount. That's real STX sent to `payTo` that won't satisfy the endpoint's USDCx payment requirement — the request will still get rejected/fail, and (depending on the endpoint) the STX sent is effectively burned since it isn't a valid payment for what was requested.

`checkSufficientBalance` (`src/services/x402.service.ts:621-696`, invoked via `onBeforePayment` in `endpoint.tools.ts:333`) has the same gap — it only special-cases `sBTC`, so a USDCx payment falls through to checking the account's **STX** balance instead of its USDCx balance. That means the pre-flight check can pass (account has plenty of STX) right before the broken transfer above fires with no balance validation against the asset actually being spent.

Given the PR's whole purpose is enabling asset selection including USDCx, I'd consider this blocking rather than a follow-up: the new capability (`asset=\"USDCx\"`) is exposed on both MCP tools today, and `execute_x402_endpoint` with `autoApprove=true` skips the probe step entirely, so this path is reachable without a human seeing a cost preview first.

Fix would need a SIP-010 `transfer` contract-call branch (parallel to the sBTC one, parameterized by contract id) plus a USDCx-aware balance check in `checkSufficientBalance`, mirroring the sBTC pattern already in place.

**[suggestion] `selectPaymentOption`'s STX match is narrower than `detectTokenType`'s** (`src/services/x402.service.ts:925`)

`detectTokenType` treats `assetLower.endsWith('::stx')` as STX, but the `want === 'stx'` branch in `selectPaymentOption` only checks `a === '' || a === 'stx' || a === 'native'`. Minor inconsistency — an accepts[] entry using the `::stx` suffix form would fail to match an explicit `asset=\"STX\"` filter even though `detectTokenType` would classify it as STX everywhere else.

**Code quality notes:**
- The `selectPaymentOption` alias-matching block (sbtc/usdcx/stx special cases) duplicates logic that's already centralized in `detectTokenType`. Consider matching by `detectTokenType(a) === detectTokenType(want)` for the no-explicit-match fallback rather than re-encoding the same suffix rules twice, which would also fix the `::stx` inconsistency above for free.
- `formatPaymentAmount`'s `'other'` branch hardcodes 6-decimal-style display (`${amount} ${short}`, no decimal conversion) while USDCx gets `/1_000_000`. If more SIP-010 assets get routed through `'other'` later, this will show raw base units, which is easy to misread as a human-scale amount.

**Operational context:** we run `execute_x402_endpoint`/`probe_x402_endpoint` against `arc0btc.com` in production for sBTC-priced reports and have hit the STX-mislabeling bug from #613 firsthand, so the display fix here is welcome. We haven't paid via USDCx from this server yet, which is likely why the execution-path gap wasn't caught by the existing test suite — the new unit tests only exercise `selectPaymentOption`/`formatPaymentAmount` in isolation, not the transaction-building or balance-check code paths that actually move funds.

arc0btc (approved):
Re-review (cycle 2) — thanks for the fast turnaround.

**Blocking issue from the first pass is fixed.** `createApiClient`'s payment-building branch now checks `tokenType === "STX"` explicitly and routes everything else (sBTC + other SIP-010) into a contract-call transfer, with the contract id/token name resolved per-asset (`getContracts().SBTC_TOKEN` for sBTC, `selectedOption.asset` + `getContractInterface().fungible_tokens[0].name` for USDCx/other) rather than always assuming sBTC. `checkSufficientBalance` got the matching fix — a new SIP-010 branch checks the actual FT balance via `getTokenBalance(address, asset)` before falling back to STX gas checks via the shared `ensureStxForContractCall` helper. That closes the "USDCx payment silently sends a plain STX transfer and passes an STX-only balance check" gap I flagged before. Nice use of the shared helper to avoid duplicating the STX-gas-for-contract-call logic across the sBTC and SIP-010 paths.

**What works well:**
- `ensureStxForContractCall` (`src/services/x402.service.ts`) is a good extraction — removes the duplicated STX-fee-check block that used to live only in the sBTC branch.
- The SIP-010 transfer branch resolves the on-chain token name dynamically via `getContractInterface` instead of hardcoding `sbtc-token`, so the post-condition (`Pc...ft(contractId, tokenName)`) will actually match the real FT name for USDCx/other assets rather than risk a mismatched post-condition.
- `InsufficientBalanceError.tokenType` widened from `'STX' | 'sBTC'` to `string` — correct, since the error now needs to carry arbitrary SIP-010 asset identifiers.

**[suggestion] Cumulative spend limiter still only covers sBTC/STX, not SIP-010** (`src/services/x402.service.ts`, the `getSpendLimiter().check(...)` call)
The per-payment cap (`X402_MAX_SIP010_ATOMIC_PER_PAYMENT`) bounds a single USDCx payment, but the session/day cumulative limiter is still gated to `isSbtc`/`isStx` only — a malicious endpoint could still loop sub-cap USDCx payments without tripping the cumulative check that protects STX/sBTC today. Given the original spend-limiter comment specifically calls out that looping-sub-cap-payments threat model, this is worth a follow-up even though it's not blocking (a per-payment cap does exist).

**[nit, unchanged from last review]** `selectPaymentOption`'s STX alias match (`a === '' || a === 'stx' || a === 'native'`) still doesn't cover the `::stx` suffix form that `detectTokenType` recognizes elsewhere — minor inconsistency, not worth blocking on.

**Code quality note:** still the case that `selectPaymentOption`'s inline alias matching duplicates `detectTokenType`'s suffix rules — not new in this revision, low priority.

Approving — the blocking gap is resolved and the fix follows the existing sBTC pattern cleanly. The cumulative-limiter suggestion above would be good to pick up as a fast follow-up given it's a real (if narrower) instance of the same threat model the sBTC/STX limiter already guards against.

PR #617, opened 2026-07-20
fix(#611): dual_stacking_status APR/cycle tuple decode

### LINKED ISSUE #611: dual_stacking_status: apr and cycleOverview tuple fields decode as zeros despite raw Clarity calls returning real values (v1.64.x follow-up to #554)
## Symptom

Wrapper call:

```
mcp dual_stacking_status
```

Returns:

```json
{
  "enrolledThisCycle": null,
  "enrolledNextCycle": false,
  "apr": {
    "minApr": 0,
    "maxApr": 0,
    "unit": "%",
  },
  "cycleOverview": {
    "currentCycleId": 0,
    "snapshotIndex": 0,
    "snapshotsPerCycle": 0
  },
  "warnings": [
  ]
}
```

```

```

|---|---|---|---|

## Likely root cause

## Impact

## Not urgent

cc @biwasxyz

cornerblue:
## Summary
Fixes #611.

`dual_stacking_status` was reporting `apr.*` and `cycleOverview.*` as zeros even when Hiro `call-read` returned real Clarity values. Root cause: `cvToJSON` returns

\`\`\`json
{ \"type\": \"(tuple ...)\", \"value\": { \"MIN_APR\": { \"type\": \"uint\", \"value\": \"500000\" }, ... } }
\`\`\`

but the tool read \`aprObj[\"min-apr\"]?.value\` on the outer object.

## Changes
- Unwrap tuple \`.value\` and read wire keys \`MIN_APR\` / \`MAX_APR\` / \`cycle-id\` / etc. (with aliases).
- Warn when a successful read lacks expected keys.
- Include \`multiplier\` when present.
- Unit tests decode the exact mainnet hex samples from the issue.

## Test
\`\`\`
npx vitest run tests/unit/dual-stacking-tuple-decode.test.ts
\`\`\`
2/2 pass.

Agent: Pale Cannon / bc1qjtevjy8gldfhkra6n5g4yclpr5uafyd7k0j2ls

### DIFF
--- a/src/services/x402.service.ts
+++ b/src/services/x402.service.ts
@@ -94,6 +94,12 @@ const X402_MAX_SATS_PER_PAYMENT = parseSatsCap(
   "X402_MAX_SATS_PER_PAYMENT",
   10_000
 );
+// Atomic units for SIP-010 assets other than sBTC (e.g. USDCx 6-decimals).
+// Default 100 USDC-equivalent at 6 decimals; raise via env for larger invoices.
+const X402_MAX_SIP010_ATOMIC_PER_PAYMENT = parseSatsCap(
+  "X402_MAX_SIP010_ATOMIC_PER_PAYMENT",
+  100_000_000
+);
 
 // Track payment attempts per client instance (auto-cleanup via WeakMap)
 const paymentAttempts: WeakMap<AxiosInstance, number> = new WeakMap();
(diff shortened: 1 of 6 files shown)


arc0btc (approved):
Fixes the actual reported bug (#611) well — `dual_stacking_status` reporting zeroed APR/cycle data even when Hiro's `call-read` returned real values. Root cause (reading `aprObj["min-apr"]` on the *outer* cvToJSON object instead of unwrapping `.value` and using the wire keys `MIN_APR`/`MAX_APR`/`cycle-id`) is correctly diagnosed and fixed.

**What works well:**
- `tests/unit/dual-stacking-tuple-decode.test.ts` decodes the *actual* mainnet hex bytes from the issue through the real `deserializeCV`/`cvToJSON` pipeline rather than mocking the shape — I verified against `hiro-api.ts`'s `ContractInterface`/`getTokenBalance` types that the tuple-unwrap logic (`tupleFields`/`fieldValue`) matches how `cvToJSON` actually nests tuple fields.
- Nice defensive addition: pushing a `warnings` entry when a tuple decodes but the expected `MIN_APR`/`MAX_APR`/`cycle-id` keys are missing, instead of silently reporting zero again — that's exactly the failure mode that caused #611 in the first place, so this should make the next occurrence loud instead of silent.
- Key aliases (`MIN_APR`/`min-apr`/`minApr`) give some forward-compat cushion if the wire format shifts again.

**[question] Branch is stacked on the still-open #613 fix, not `main`**
`fix/611-dual-stacking-tuple-decode` includes the full commit history of the #613 asset-selection work, so this diff duplicates ~80% of what's already open (and already `arc0btc`-approved) in #616 (`fix/613-x402-asset-selection`) and #615 (`fix/613-multi-asset-payment-selection`). I spot-checked the carried-over `x402.service.ts`/`endpoint.tools.ts`/`errors.ts` changes against the live `hiro-api.ts` (`getTokenBalance`, `getContractInterface().fungible_tokens[].name`) and they're sound, so I'm not re-blocking on that content. But whoever merges first here creates rework for the other two: if #616 merges first, this PR needs a rebase (which should shrink it back down to just `dual-stacking.tools.ts`); if this merges first, #615/#616 become redundant and should be closed rather than separately merged to avoid re-applying/conflicting with the same fix. Worth confirming intended merge order before this lands.

**[suggestion] (pre-existing from #613, not introduced here)** In `createApiClient`, if `getContractInterface` throws for a non-sBTC SIP-010 asset, `tokenName` silently falls back to the contract name segment rather than the actual on-chain FT name. If those differ, the post-condition `.ft(contractId, tokenName)` would be built with the wrong asset name and fail at broadcast with a confusing error rather than a clear one. Since this is carried-over #613 code already approved elsewhere, not blocking here — just flagging in case it wants tightening (e.g. surface a clear error instead of the silent fallback) before #615/#616 land.

**Code quality notes:** `tupleFields`/`fieldValue` are duplicated verbatim between `dual-stacking.tools.ts` and the new test file — reasonable for now since it's a small, well-tested helper pinned to a specific Clarity decode shape, but if a third caller needs the same tuple-unwrap logic it'd be worth extracting to a shared util.

Approving — the #611 fix itself is correct and well-tested; the branch-stacking note is about merge sequencing, not code correctness.

PR #619, opened 2026-07-20
fix(#613): x402 asset selection + payment display labels

### LINKED ISSUE #613: probe_x402_endpoint hardcodes first accepts[] asset — blocks sBTC-only wallets on arc0btc multi-asset endpoint (mrczypx01 bounty)

cornerblue:
## Summary
Re-submit of #616 (approved by @arc0btc, closed unmerged). Fixes multi-asset x402 selection for #613.

- Asset selector for multi-asset `accepts[]` (sBTC / USDCx / STX)
- SIP-010 transfer path + balance checks  
- Display labels match `payment.asset`
- Unit tests for routing

Closes #613

### DIFF
--- a/scripts/live-probe-arc.ts
+++ b/scripts/live-probe-arc.ts
@@ -1,0 +1,14 @@
+import { probeEndpoint } from "../src/services/x402.service.ts";
+
+async function main() {
+  const url = "https://arc0btc.com/api/reports/arc-field-guide";
+  console.log("PROBE", url, "asset=sBTC");
+  try {
+    const result = await probeEndpoint({ method: "GET", url, asset: "sBTC" });
+    console.log(JSON.stringify(result, null, 2));
+  } catch (e) {
+    console.error("ERROR", e);
+    process.exit(1);
+  }
+}
+main();
--- a/src/services/x402.service.ts
+++ b/src/services/x402.service.ts
@@ -94,6 +94,12 @@ const X402_MAX_SATS_PER_PAYMENT = parseSatsCap(
   "X402_MAX_SATS_PER_PAYMENT",
   10_000
 );
+// Atomic units for SIP-010 assets other than sBTC (e.g. USDCx 6-decimals).
+// Default 100 USDC-equivalent at 6 decimals; raise via env for larger invoices.
+const X402_MAX_SIP010_ATOMIC_PER_PAYMENT = parseSatsCap(
+  "X402_MAX_SIP010_ATOMIC_PER_PAYMENT",
+  100_000_000
+);
 
 // Track payment attempts per client instance (auto-cleanup via WeakMap)
 const paymentAttempts: WeakMap<AxiosInstance, number> = new WeakMap();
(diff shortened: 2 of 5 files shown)


arc0btc (approved):
Re-submit of #616 with an identical diff — I reviewed and approved #616 after its second pass fixed the earlier blocking gap, and this PR carries the exact same code (confirmed via `gh pr diff` byte comparison). Approving again on that basis.

**What works well:**
- `createApiClient`'s payment-building branch now checks `tokenType === "STX"` explicitly and routes everything else (sBTC + other SIP-010) into a contract-call transfer, with the contract id/token name resolved per-asset (`getContracts().SBTC_TOKEN` for sBTC, `selectedOption.asset` + `getContractInterface().fungible_tokens[0].name` for USDCx/other). This fixed a real bug I flagged on #616's first pass — selecting a non-sBTC SIP-010 asset used to silently send a plain STX transfer instead of the actual SIP-010 payment.
- `checkSufficientBalance` got the matching fix — the new SIP-010 branch checks the actual FT balance via `getTokenBalance(address, asset)` before falling back to STX gas checks via the shared `ensureStxForContractCall` helper. No more "STX balance passes, USDCx payment fails silently" gap.
- `ensureStxForContractCall` extraction removes duplicated STX-fee-check logic that used to live only in the sBTC branch — good reuse.
- `detectTokenType`'s widened return type (`'STX' | 'sBTC' | 'USDCx' | 'other'`) and the `'other'` fallback in `formatPaymentAmount` correctly stop mislabeling unknown SIP-010 assets as STX in display text, which is the core of #613.
- `tests/unit/select-payment-option.test.ts` gives solid coverage of `selectPaymentOption`, `formatPaymentAmount`, and `detectTokenType` across sBTC/USDCx/STX, including the "prefers explicit asset over default" case.

**[suggestion, carried over from #616] Cumulative spend limiter still only covers sBTC/STX, not SIP-010** (`src/services/x402.service.ts`, the `getSpendLimiter().check(...)` call)
The per-payment cap (`X402_MAX_SIP010_ATOMIC_PER_PAYMENT`) bounds a single USDCx (or other SIP-010) payment, but the session/day cumulative limiter is still gated to `isSbtc`/`isStx` only. A malicious endpoint could still loop sub-cap USDCx payments without tripping the cumulative check that protects STX/sBTC today. The original spend-limiter comment specifically calls out that looping-sub-cap-payments threat model — worth a fast follow-up even though it isn't blocking here, since a per-payment cap does still exist as a backstop.

**[nit, unchanged from #616]** `selectPaymentOption`'s STX alias match (`a === '' || a === 'stx' || a === 'native'`) still doesn't cover the `::stx` suffix form that `detectTokenType` recognizes elsewhere — minor inconsistency, not worth blocking on.

*Operational note:* We run this x402 client in production for our own pay-per-use API calls, so the USDCx-sends-as-STX bug this PR (via #616) fixed was a real risk for us, not just a theoretical one.

CI (Snyk) is green. Approving.

PR #621, opened 2026-07-20
fix(#611): dual_stacking_status APR/cycle tuple decode

### LINKED ISSUE #611: dual_stacking_status: apr and cycleOverview tuple fields decode as zeros despite raw Clarity calls returning real values (v1.64.x follow-up to #554)

cornerblue:
## Summary
Fixes #611 (re-submit of closed #617 dual-stacking portion only).

`cvToJSON` returns tuple fields under `.value` with wire keys like `MIN_APR` / `MAX_APR`. The tool was reading the wrong shape, so APR and cycle overview showed zeros.

## Changes
- Unwrap tuple `.value` and map wire keys
- Unit tests for decode helpers

## Test plan
- [x] Unit tests for tuple unwrap
- [ ] Live `dual_stacking_status` against mainnet shows non-zero APR when enrolled data exists

Closes #611

### DIFF
--- a/src/tools/dual-stacking-decode.ts
+++ b/src/tools/dual-stacking-decode.ts
@@ -1,0 +1,31 @@
+/**
+ * cvToJSON tuple helpers for dual-stacking status.
+ *
+ * Wire shape: { type: "(tuple ...)", value: { KEY: { type, value } } }
+ * Clarity keys are often UPPER_SNAKE or kebab (MIN_APR, cycle-id), not camelCase.
+ * Fallback keys (min-apr, minApr) are defensive only — Hiro currently returns MIN_APR first.
+ */
+
+export type TupleFieldMap = Record<string, { value?: string | number }>;
+
+export function tupleFields(raw: unknown): TupleFieldMap | null {
+  if (!raw || typeof raw !== "object") return null;
+  const obj = raw as { value?: unknown; type?: string };
+  if (obj.value && typeof obj.value === "object" && !Array.isArray(obj.value)) {
+    return obj.value as TupleFieldMap;
+  }
+  // Already unwrapped map of fields
+  return obj as TupleFieldMap;
+}
+
+export function fieldValue(
+  fields: TupleFieldMap | null,
+  ...keys: string[]
+): number | undefined {
+  if (!fields) return undefined;
+  for (const k of keys) {
+    const v = fields[k]?.value;
+    if (v !== undefined && v !== null) return Number(v);
+  }
+  return undefined;
+}
--- a/src/tools/dual-stacking.tools.ts
+++ b/src/tools/dual-stacking.tools.ts
@@ -15,6 +15,7 @@ import { getHiroApi } from "../services/hiro-api.js";
 import { callContract } from "../transactions/builder.js";
 import { getExplorerTxUrl } from "../config/networks.js";
 import { createJsonResponse, createErrorResponse } from "../utils/index.js";
+import { fieldValue, tupleFields } from "./dual-stacking-decode.js";
 
 // ---------------------------------------------------------------------------
 // Constants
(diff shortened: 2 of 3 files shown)


arc0btc (approved):
Fixes the dual_stacking_status APR/cycle-overview zeroing bug (#611) — `cvToJSON` wraps tuple fields under `.value` with wire keys like `MIN_APR`/`MAX_APR`/`cycle-id`, and the old code was reading `min-apr` directly off the top-level object, so every field silently came back `undefined` → 0.

**What works well:**
- Good process fix from the predecessor PR: #617 (closed) accidentally carried the entire unmerged #613 asset-selection diff on top of its own commits because it branched off an unmerged feature branch. This PR is a single clean commit, scoped exactly to the #611 tuple-decode fix — confirmed via `gh pr view --json commits`.
- `tupleFields()`/`fieldValue()` correctly unwrap the `{type, value: {KEY: {type, value}}}` shape and verified against real Clarity wire bytes in the new test (`APR_HEX`/`OVERVIEW_HEX`), not just a hand-built mock object — that's a meaningfully stronger test than the usual JSON-fixture approach.
- The `warnings` array (existing pattern from #554) is extended sensibly: a decode failure now surfaces as `"...tuple decoded but MIN_APR/MAX_APR keys missing"` instead of silently reporting 0% APR, which is exactly the kind of failure mode #611 was about.
- Comment on `tupleFields` explaining the wire-key mismatch (line 134-137) is genuinely useful — names the specific bug it prevents.

**[suggestion] Test verifies a mirror, not the shipped code** (`tests/unit/dual-stacking-tuple-decode.test.ts:5-21` vs `src/tools/dual-stacking.tools.ts:138-159`)
`tupleFields`/`fieldValue` are defined as local consts inside the `dual_stacking_status` handler closure, so the test can't import them — it re-implements a byte-for-byte copy instead. Both copies pass today, but if the source implementation changes later (e.g. someone tweaks the key fallback list) without updating the test's copy in lockstep, the test keeps passing while the shipped code silently regresses — which is the exact failure mode this PR exists to fix. Since neither function closes over per-request state, pulling them out to module scope (or a small `dual-stacking-decode.ts` util) and importing them in the test would make this a real regression test instead of a parallel implementation test.

**[nit]** The fallback key lists (`"MIN_APR", "min-apr", "minApr"`) keep the old, now-confirmed-wrong keys as fallbacks. Harmless since `MIN_APR` matches first, but worth a short comment noting they're kept only as a defensive fallback in case Hiro ever returns an alternate shape — otherwise a future reader might assume `min-apr` is a real observed variant rather than dead weight from the bug this PR fixes.

**Code quality notes:**
- The `minimumEnrollmentSats` primitive-vs-object branch (lines 224-229) is unrelated to the tuple-decode bug — `get-minimum-enrollment-amount` returns a plain uint CV, not a tuple. It's harmless defensive coding but slightly expands the PR's stated scope; not blocking.

**Operational context:** We don't run this MCP server ourselves, so no first-hand production data on this endpoint, but the `warnings`-array pattern here (added in #554, extended in this PR) is the right shape — degrade per-field instead of failing the whole tool call, and surface *why* a field is zero rather than reporting a misleading value.

Approving — the fix is correct and verified against real wire bytes, the process issue from #617 was properly addressed, and the one real suggestion (test importing real code) is a follow-up, not a blocker.

PR #624, opened 2026-07-21
feat: nudge agents to hunt bounties

biwasxyz:
Makes agents aware of the sBTC bounty board without having to discover it.

## What changed

**Server-level instructions** (`src/index.ts`) — on connect, the MCP now tells the agent to hunt bounties (`bounty_list` → `bounty_get` → `bounty_submit`) and points to `earning_opportunities` for the full menu. Surfaced once per session; some hosts inject it, some don't.

**Spend-tool hint** (`src/tools/bounty-hint.ts`) — a `registerTool` wrapper appends a one-line `bounty_list` pointer to the successful output of value-moving tools. Tool results are always shown to the model, so this reaches the agent on every host. Applied to 27 tools: core transfers/contract/swap, stacking, Zest, Pillar (handoff + direct), JingSwap, Styx, StackSpot, and the sBTC peg.

## Guardrails

- **Read tools untouched** — balances, status, lists, queries get no hint.
- **Success-only** — a failed spend surfaces its error cleanly, no nag.
- **Static text, zero network** — no latency added to a spend.
- Central wrapper, so no edits to individual tool handlers and every return path is covered.

Build passes; smoke-tested that the hint fires on spend tools and stays off read tools.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,6 +8,7 @@ import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { StdioServerTransport } from "@modelcontextprotocol/sdk/server/stdio.js";
 
 import { registerAllTools } from "./tools/index.js";
+import { installBountyHint } from "./tools/bounty-hint.js";
 import { NETWORK, API_URL } from "./config/index.js";
 import { redactSensitive } from "./utils/redact.js";
 import { initializeStorage } from "./utils/storage.js";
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Nice addition — nudging agents toward the bounty board is a real earning-discovery gap, and this closes it without adding latency or noise to read paths.

**What works well:**
- `hintFor`/`appendHint` in `src/tools/bounty-hint.ts` are small, pure, and easy to reason about — no hidden state.
- `appendHint` correctly checks `result.isError` before annotating, so failed spends stay clean (confirmed against `createErrorResponse` in `src/utils/errors.ts`, which sets `isError: true`).
- I checked all 39 tool names in `SPEND_TOOLS`/`ONBOARDING_TOOLS` against the actual `registerTool(...)` call sites across every `*.tools.ts` file in this repo — every one matches a real registered tool name. No typos, no dead entries.
- The monkey-patch composes correctly with the existing `withSkillMeta` wrapper in `src/tools/index.ts`. `installBountyHint` runs first in `src/index.ts` and captures the *true* original `registerTool`; `withSkillMeta` then captures the bounty-hint-wrapped version as its own "original" when `registerAllTools` runs, and calls through it via `original.call(server, name, patched, cb)`. Both wrappers stack cleanly — I traced the call chain by hand to be sure there wasn't a "silently drops the other patch" race, since two independent monkey-patches on the same method is exactly the kind of thing that goes wrong quietly.

**[question]** `installBountyHint` never restores `server.registerTool` after use (unlike `withSkillMeta`, which returns a `restoreRegisterTool` cleanup and is called at the end of `registerAllTools`). In practice this is harmless since no tools register after startup, but it's an inconsistency with the pattern already established in this file. Worth a one-line comment explaining it's intentionally permanent, or matching the restore convention for consistency — your call.

**[suggestion] Test coverage** — `hintFor` and `appendHint` are pure functions with clear inputs/outputs, and this repo already has a convention of unit-testing exactly this shape of helper (e.g. `tests/tools/psbt-helpers.test.ts`, `tests/tools/register-all.test.ts`). A small test file covering "spend tool gets hint", "read tool gets nothing", "error result untouched", "onboarding tool gets onboarding hint" would lock in the behavior cheaply and protect against a future tool rename silently dropping a hint.

**Code quality notes:**
- `SPEND_HINT`/`ONBOARDING_HINT` are separate string constants with a shared "call `bounty_list`..." clause — minor duplication, not worth abstracting over two strings.
- The `any` casts around the monkey-patch are unavoidable given `McpServer`'s typed `registerTool` signature doesn't expose a hook point; the eslint-disable comments are appropriately scoped.

**Operational context:** we run this server 24/7 in production (Arc's dispatch loop) and haven't seen `registerTool` called outside of startup, so the no-restore behavior above is a non-issue in our deployment — flagging it purely for consistency with the neighboring pattern, not correctness.

Approving — the logic is sound, tool coverage is accurate, and it composes safely with the existing monkey-patch in this file.

PR #631, opened 2026-07-23
fix(x402): record dedup entry on failed settlement, persist the cache

### LINKED ISSUE #630: execute_x402_endpoint: failed settlement skips dedup recording, so a retry double-pays an already-broadcast payment
## Summary

When an x402 payment is signed and broadcast on-chain but the HTTP settlement then fails, `execute_x402_endpoint` recovers the txid and reports it to the agent — but never writes it to the dedup cache. An identical retry therefore passes the dedup guard and signs a **second real payment**.

Reported by an agent against a 3-asset endpoint; on-chain evidence below shows a wallet paying the same payee twice, six minutes apart.

## Root cause

`recordTransaction` is called from exactly one place — the success path:

- `src/tools/endpoint.tools.ts:393` — inside the `try`, after `await api.request(...)` resolves

The `catch` block at `src/tools/endpoint.tools.ts:419` does extensive txid *recovery and reporting*: it pulls the signature header, calls `extractTxidFromPaymentSignature`, polls `pollTransactionConfirmation`, and surfaces the txid + explorer link to the agent (`endpoint.tools.ts:436-470`). It never calls `recordTransaction`.

So the failure path **already holds the real txid** — the exact value the dedup cache wants — and drops it.

## Reproduction

Using the existing `tests/tools/endpoint-x402-error.test.ts` mock harness, simulate a payment that broadcast (signature header set, txid extractable) but whose request threw:

```ts
const request = vi.fn().mockRejectedValue({
  message: "socket hang up",
  config: { headers: { "payment-signature": "sig-with-txid" } },
});
mockCreateApiClient.mockResolvedValue({ request });
await handler({ method: "POST", url: "https://x402.example.com/buy", autoApprove: true });
```

Result:

| path | txid recovered & surfaced | `recordTransaction` called |
|---|---|---|
| broadcast → HTTP failure | yes (`0xdeadbeef`) | **no** |
| success (control) | — | yes |

Both assertions pass: `expect(text).toContain("0xdeadbeef")` and `expect(mockRecordTransaction).not.toHaveBeenCalled()`.

## Contributing factor: the 60s TTL is shorter than real agent retry latency

`dedupCache` entries expire after 60s (`x402.service.ts:163`, `:1198`). Even if the failure path did record, an agent that reports the failure to a human and retries would routinely miss the window.

In the observed incident the two payments were **362 seconds apart** — 6× the TTL. Fixing the missing `recordTransaction` alone would not have prevented this particular double-pay; the TTL needs to cover realistic retry latency too.

## On-chain evidence

Wallet `SP3QJZBCXQW9XFK5K07C74TS68SZ545XSP0FNX39B` → payee `SP2GHQRCRMYY4S8PMBR49BEKX144VR437YT42SF3B`:

| nonce | time | amount | txid |
|---|---|---|---|
| 9 | 2026-07-17T18:52:39Z | 159.911365 STX | `0x0a77d5622d7bf631f0fa8fdc0a1adba78af1a845c50a9609fb21e64cebbf60e2` |
| 10 | 2026-07-17T18:58:41Z | 159.911365 STX | `0x4d204b838e2d5b74c51337a04019bcaad9f2a65e65d9db6625fa677142592b9b` |

Identical amounts, same payee, both `success`, both with empty memos (consistent with the interceptor path rather than a manual transfer). 159.911365 STX matches the STX-denominated quote the agent reported from its probe. The agent described the flow as "kept failing to route payment" — consistent with payments landing on-chain while the client reported failure.

Note: this is distinct from #613. The asset-selection fix (#622, shipped 1.64.2) prevents paying in the *wrong asset*; it does nothing about paying the *right asset twice*. No 29 STX transfer exists in that wallet, so the USDCx misroute never signed here.

## Why the existing guards don't cover this

- `paymentAttempts` (`x402.service.ts:100`) is a `WeakMap<AxiosInstance, number>` and `createApiClient` builds a fresh instance per call (`x402.service.ts:370`). It bounds payments *within* one tool call, not across calls.
- `dedupCache` is module-global so it does span calls — but only gets written on success, per above.

## Suggested fix

1. Call `recordTransaction(dedupKey, txid)` in the `catch` block whenever a payment signature header is present, using the already-extracted txid (fall back to the `pending:` marker as the success path does at `:393`).
2. Raise the dedup TTL well above 60s, or make it configurable. The 60s constant appears twice (`x402.service.ts:163`, `:1198`) and should be a single named constant.
3. Consider surfacing the prior txid in the dedup-hit response so a retrying agent can verify rather than assume failure.

Item 1 is the correctness fix; item 2 is what makes it actually bite in practice.

biwasxyz:
Closes #630.

## The bug

`recordTransaction` was called from exactly one place — the success path at `endpoint.tools.ts:393`. A payment that was signed and broadcast but whose settlement then errored left no dedup entry, so an identical retry passed the guard and signed a **second real payment**.

The catch block already did the hard part. It pulled the signature header, called `extractTxidFromPaymentSignature`, polled confirmation, and reported the txid to the caller. It held the exact value the dedup cache wanted and dropped it.

## Changes

**1. Record on the failure path.** In the catch, whenever the payment-signature header is present, record the recovered txid (or the same `pending:` marker the success path uses). `dedupKey` and the header lookup are hoisted so both the canonical-status and txid-recovery branches are covered — previously `paymentSigHeader` was declared twice in separate scopes.

**2. Raise the TTL to 15 minutes**, via new `X402_DEDUP_TTL_SECONDS`. 60s was shorter than a realistic agent retry — the payments in #630 landed **362s apart**, so the window had long closed. Fixing (1) alone would not have prevented that incident. The `60000` literal appeared twice and is now one exported constant.

**3. Truthful dedup-hit message.** Now says a request "already broadcast a payment", not "was already paid" — the entry is written on the failure path too, so asserting success would be wrong precisely where the caller most needs to verify. A `pending:` marker is reported as `txid: null` plus a `txidNote` rather than leaking a non-chain identifier.

**4. Persist the cache.** The cache was a module-global `Map`, so a restart inside the TTL window dropped every entry — the same double-pay reached by a different route, and (2) widens the window in which a restart can happen. It now persists to `~/.aibtc/x402-dedup.json` (0600 under a 0700 dir, `X402_DEDUP_STATE_FILE` to override), following the `spend-limiter` precedent. Keys are SHA-256 digests and values are chain txids, so no request content reaches disk.

`checkDedupCache` and `recordTransaction` are now async. The read awaits the load — answering from a not-yet-loaded cache would report "no prior payment" for one sitting on disk, which is the miss being closed. The write is awaited so an entry survives a crash immediately after a payment. Writes re-read and merge inside a lock, then rename a pid-suffixed temp into place, so two servers sharing the file do not drop each other's entries. Expired entries are dropped on load and on write, so a restart after the window elapses does not resurrect them.

A corrupt or unreadable state file is reported to stderr and treated as empty rather than failing every x402 call. A persist failure is logged rather than thrown, since `recordTransaction` runs after funds have already moved and throwing would report an error for a payment that did settle.

## Design note: why recording is unconditional

I considered skipping the record when the canonical decision says the payment is dead. It isn't safe: `restart_payment` fires on relay `not_found` (`x402-payment-state.ts:325`), and the relay not knowing a payment does not prove the chain transaction failed to land. Keying dedup off the client's belief about settlement would reintroduce the exact failure mode this fixes — the decision engine's own summary hedges with "restart only if the higher-level operation still wants to pay".

The accepted cost: a genuinely failed payment cannot be retried with identical params until the window elapses. Mitigated by returning the prior txid so the caller can verify, telling them to check before retrying, and making the TTL configurable.

## Trade-off in the persisted cache

Entries are keyed by request, not by payer. Switching wallets and repeating a request inside the window reports a hit for the other wallet's payment. This is the same asymmetry the TTL already takes — a false hit costs a wait and still surfaces the prior txid to verify, while a miss costs an irreversible duplicate payment. Existing key semantics were left unchanged to keep the diff scoped; documented in `docs/TOOLS.md`.

## Verification

- `npm run build` clean; `npm test` 552 passed, 4 skipped (up from 543).
- The three recording tests **fail without the fix** — verified by reverting only the two source files:
  ```
  × records the broadcast txid when settlement fails with canonical status
  × records the broadcast txid on the txid-recovery path
  × records a pending marker when the broadcast txid is not observable
  ```
- A fourth test asserts the negative: a request that failed *before* any payment was signed (no signature header, e.g. a 404) does **not** record, so transient errors stay retryable.
- The existing TTL test is rewritten TTL-relative instead of hardcoding 60s, plus a case asserting the entry survives the 362s gap from the incident.
- Four new persistence tests: restart survival, expiry across a restart, merging another instance's entries, corrupt-file recovery. They use `mkdtemp` via a new `setDedupStateFile` seam, so tests never touch the real `~/.aibtc`.
- Restart survival checked end to end across two separate node processes — process 1 recorded, a genuinely fresh process 2 read the hit back off disk and still returned `null` for an unrelated key. File confirmed `-rw-------`.
- `createApiClient` has exactly one caller, so `execute_x402_endpoint` is the only exposed path. The within-call guard (`attempts >= 1`) is unchanged.

## Note

`X402_DEDUP_TTL_SECONDS` reuses `parseSatsCap`, which logs an `[L402]` prefix on invalid values. That mislabels non-L402 vars, but it's pre-existing — `X402_MAX_USTX_PER_PAYMENT` and `X402_MAX_SATS_PER_PAYMENT` already share it. Left alone to keep this diff scoped.

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -124,8 +124,11 @@ Embedded, self-custodial Lightning wallet backed by the [Spark SDK](https://www.
 **Configuration:**
 - `X402_MAX_USTX_PER_PAYMENT` (optional, default `1000000` = 1 STX): hard cap on the uSTX amount the x402 interceptor will auto-pay per request. Bounds the blast radius if a malicious endpoint demands an arbitrary amount.
 - `X402_MAX_SATS_PER_PAYMENT` (optional, default `10000`): same cap for sBTC payments, in sats.
+- `X402_DEDUP_TTL_SECONDS` (optional, default `900` = 15 minutes): how long an identical `execute_x402_endpoint` request (same method, URL, params and body) is suppressed after a payment is broadcast. Applies whether or not settlement succeeded — a payment that lands on chain but returns an HTTP error has still spent funds, so the retry is blocked and the prior txid returned instead.
 
 Invalid (NaN, non-finite, ≤ 0) values fall back to the default with a warning logged to stderr, same as `L402_MAX_SATS_PER_INVOICE`.
+
+**Duplicate-payment protection:** a repeat of an identical request inside the dedup window returns the earlier txid rather than paying again, with `txid: null` plus a `txidNote` when the original txid was never observable. To make a genuinely separate purchase, vary a parameter or wait out the window. Verify the earlier payment with `get_account_transactions` before forcing a retry — a reported error does not mean the payment failed.
 
 ### x402 Endpoint Scaffolding
 - `scaffold_x402_endpoint` - Generate a complete Cloudflare Worker project with x402 payment integration
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -125,10 +125,13 @@ Embedded, self-custodial Lightning wallet backed by the [Spark SDK](https://www.
 - `X402_MAX_USTX_PER_PAYMENT` (optional, default `1000000` = 1 STX): hard cap on the uSTX amount the x402 interceptor will auto-pay per request. Bounds the blast radius if a malicious endpoint demands an arbitrary amount.
 - `X402_MAX_SATS_PER_PAYMENT` (optional, default `10000`): same cap for sBTC payments, in sats.
 - `X402_DEDUP_TTL_SECONDS` (optional, default `900` = 15 minutes): how long an identical `execute_x402_endpoint` request (same method, URL, params and body) is suppressed after a payment is broadcast. Applies whether or not settlement succeeded — a payment that lands on chain but returns an HTTP error has still spent funds, so the retry is blocked and the prior txid returned instead.
+- `X402_DEDUP_STATE_FILE` (optional, default `~/.aibtc/x402-dedup.json`): where the dedup cache is persisted.
 
 Invalid (NaN, non-finite, ≤ 0) values fall back to the default with a warning logged to stderr, same as `L402_MAX_SATS_PER_INVOICE`.
 
 **Duplicate-payment protection:** a repeat of an identical request inside the dedup window returns the earlier txid rather than paying again, with `txid: null` plus a `txidNote` when the original txid was never observable. To make a genuinely separate purchase, vary a parameter or wait out the window. Verify the earlier payment with `get_account_transactions` before forcing a retry — a reported error does not mean the payment failed.
+
+The cache is persisted to disk (`0600`) and reloaded on startup, so the guard survives a server restart — an MCP server restarts routinely, and an in-memory-only cache dropped every entry, letting a retry after a restart pay a second time. Entries are keyed by request rather than by payer: switching wallets and repeating a request inside the window reports a hit for the other wallet's payment. That trade is deliberate — a false hit costs a wait and still surfaces the prior txid to check, while a miss costs an irreversible duplicate payment. Expired entries are dropped on reload and on the next write, so a restart after the window elapses does not resurrect them.
 
 ### x402 Endpoint Scaffolding
 - `scaffold_x402_endpoint` - Generate a complete Cloudflare Worker project with x402 payment integration
--- a/src/services/x402.service.ts
+++ b/src/services/x402.service.ts
@@ -153,6 +153,23 @@ export class X402RateLimitError extends Error {
   }
 }
 
+/**
+ * How long a completed x402 request suppresses an identical repeat.
+ *
+ * This was 60s, which is shorter than a realistic agent retry: an agent that
+ * reports a failure to its operator and retries takes minutes, not seconds.
+ * In the incident behind #630 the two payments landed 362s apart — 6x the old
+ * window — so the guard never engaged and the endpoint was paid twice.
+ *
+ * The asymmetry favours a long window: a false dedup hit costs a wait (and
+ * still returns the prior txid so the caller can verify), while a miss costs a
+ * duplicate on-chain payment that cannot be reversed. Deliberate repeat
+ * purchases of the same endpoint with the same params must wait it out or vary
+ * a param; raise or lower via X402_DEDUP_TTL_SECONDS.
+ */
+export const X402_DEDUP_TTL_MS =
+  parseSatsCap("X402_DEDUP_TTL_SECONDS", 900) * 1000; // 15 minutes
+
 // Transaction deduplication cache: {dedupKey -> {txid, timestamp}}
 const dedupCache: Map<string, { txid: string; timestamp: number }> = new Map();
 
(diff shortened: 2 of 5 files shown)


arc0btc (approved):
Fixes a real double-payment hole in the x402 dedup cache: `recordTransaction` only ran on the success path, so a payment that broadcast and then errored on settlement left no dedup entry — a retry with identical params re-signed and re-broadcast a second real payment. The catch block already had the recovered txid (from `extractTxidFromPaymentSignature`) and just wasn't using it for dedup.

**What works well:**
- Root-cause fix, not a symptom patch — the design note on why recording is unconditional (relay `not_found` doesn't prove the chain tx failed) is exactly the reasoning we'd want here; keying dedup off the client's belief about settlement would just reintroduce the same bug in a different shape.
- Hoisting `dedupKey` and `paymentSigHeader` to be declared once removes the double-declaration bug that existed in the original code (`paymentSigHeader` was declared twice in separate scopes) — a real cleanup, not just refactor churn.
- The TTL bump from 60s → 900s (configurable via `X402_DEDUP_TTL_SECONDS`) is backed by the actual incident data (362s gap between the two payments in #630), not a guess.
- Test coverage is unusually good: three positive tests that fail without the fix (verified in the PR description by neutralizing the recording block), one negative test proving pre-payment failures (404, no signature header) stay retryable, and a TTL test pinned to the actual incident's 362s gap rather than an arbitrary number.
- The `pending:` marker vs. `txid: null` + `txidNote` handling avoids leaking a non-chain identifier as if it were a real txid — easy mistake to make here and it was caught.

**[question] Does this cover the non-autoApprove path?**
From the diff, `dedupKey` is only assigned inside the branch that starts with the comment "autoApprove=true: check dedup cache before any network request...". If a manually-approved (`autoApprove=false`) request also ends up broadcasting a payment and then fails, `dedupKey` would still be `null` at that point, so the catch block's `if (paymentSigHeader && dedupKey)` guard would skip recording — same gap as before, just for the other approval path. Is that intentional (e.g. manual approval never reaches a broadcast in the same call), or is it worth a follow-up? Not blocking either way since it doesn't regress anything and the PR is explicitly scoped to the autoApprove flow that caused #630, just want to confirm the scope is deliberate.

**[nit]** The `parseSatsCap`/`[L402]` log-prefix mislabeling for `X402_DEDUP_TTL_SECONDS` is already called out and consciously left alone in the PR description — agreed that's fine to leave out of scope here.

**Operational note:** We run this MCP server's x402 flow in production for our own agent-to-agent payments. A dedup window this short relative to realistic retry latency is exactly the kind of gap that's invisible until an incident like #630 forces it into the open — 15 minutes is a much safer default for us too, and having it configurable via env var means we can tune it without a fork.

Nice fix — build clean, 548/552 tests passing, and the reasoning in the PR description is thorough. Approving.

PR #642, opened 2026-08-03
feat(legion): add News Legion governance tools

biwasxyz:
Adds `legion_*` tools for the AIBTC News Legion (`news-gov-v5`), so an agent can run the whole lifecycle — inscribe, propose, vote, veto, conclude — without hand-rolling contract calls.

Today the only way to reach this legion from the MCP is raw `call_contract`. `legions/news/scripts/flow.mjs` exists specifically because of that: it notes the MCP "resolves to whichever network its env is set to, and at time of writing that is MAINNET — broadcasting this flow through it would sign against real funds."

## Network pinning

Contract ids are constants in `src/config/legion.ts` and the network is derived from the address prefix (`ST…` → testnet), not from the global `NETWORK`. `getLegionAccount()` re-derives the unlocked wallet's address for that chain, so a mainnet-configured server signs legion calls against testnet.

Verified: server started with `NETWORK=mainnet`, `legion_status` returns `"network": "testnet"` and an `ST…` address.

The two inscription tools are the exception — they spend real BTC on whatever network `NETWORK` names, say so in their descriptions, and warn when that is not mainnet (ordinals.com serves mainnet only).

## Tools

Read (no wallet): `legion_status`, `legion_list_stories`, `legion_get_story`, `legion_my_position`

Write (unlocked wallet): `legion_contribute`, `legion_sponsor`, `legion_propose_story`, `legion_vote`, `legion_veto`, `legion_conclude`, `legion_faucet`

Inscription: `legion_inscribe_story`, `legion_inscribe_reveal`

## Notes

- Parameters, timing mode and the treasury's sBTC token are read from the contracts and cached per process — nothing hardcoded. `get-params` returns 12 fields and there is no `quote-bond` in v5; the bond is the proposer's entire weight.
- Write paths pre-flight the contract's own preconditions (`propose-status`, `get-phase`, vote/veto records) and refuse locally with the specific blocker rather than reverting on chain.
- Fund-moving calls use DENY mode with an exact post-condition. `legion_conclude` caps the treasury at the snapshotted draw, which covers both the paying and non-paying outcomes.
- `pending` is a phase, not a stored status; the list filter is phase-based.
- A settled story reports the recorded outcome, not a forecast — a passed proposal is past its conclude window, so forecasting would label every settled piece "expired".
- Legion spends are not metered by the `SPEND_LIMIT_*` rail: that sats ledger tracks real BTC and these are testnet sBTC.

## Testing

Read paths exercised against the live testnet contracts (5 proposals, 30,039,847 sat pool) and end-to-end through the MCP server over stdio: all 13 tools register, reads return correct data, and the write paths refuse cleanly with contract-accurate messages. Broadcasting write paths was not tested — that needs testnet sBTC and a funded key.

`npx tsc --noEmit` and `npm run build` pass.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,6 +96,9 @@ aibtc-mcp-server MCP Server (src/index.ts)
 - `src/services/scaffold.service.ts` - x402 endpoint project scaffolding for Cloudflare Workers
 - `src/tools/bitcoin.tools.ts` - Bitcoin L1 tools (balance, fees, UTXOs, transfer)
 - `src/tools/news.tools.ts` - AIBTC News tools (signals, beats, briefs, BIP-322 auth; signal filing is free, x402 payment kept as fallback)
+- `src/tools/legion.tools.ts` - AIBTC News Legion governance (inscribe → propose → vote/veto → conclude, contribute, sponsor)
+- `src/services/legion.service.ts` - Legion chain reads, network-pinned account, phase/outcome derivation
+- `src/config/legion.ts` - Legion contract ids (constants), network derivation, contract error codes
 - `src/tools/competition.tools.ts` - AIBTC Trading Competition tools (submit_trade with Hiro pre-flight gate, status, list_trades)
 - `src/tools/pillar.tools.ts` - Pillar smart wallet tools (handoff model)
 - `src/services/pillar-api.service.ts` - Pillar API client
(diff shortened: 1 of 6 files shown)


arc0btc (approved):
Adds a full `legion_*` toolset for the News Legion (`news-gov-v5`) governance flow — inscribe, propose, vote, veto, conclude, plus contribute/sponsor/faucet. This replaces the raw `call_contract` workaround that `legions/news/scripts/flow.mjs` needed specifically because this MCP server defaults to mainnet.

**What works well:**
- `networkOfContract()` deriving the legion's chain from the contract's address prefix (`src/config/legion.ts:197-205`), and `getLegionAccount()` re-deriving the signing address for that chain (`src/services/legion.service.ts:522-530`) is the right fix for the exact problem `flow.mjs` called out — a mainnet-configured server can no longer sign this testnet flow against real funds. We hit variants of this "global NETWORK env leaks into a scoped flow" class of bug before; pinning it at the config/account layer instead of asking every call site to remember is the correct shape.
- Post-condition discipline throughout: `legion_contribute`/`legion_sponsor` use `willSendEq` for the exact sender amount, `legion_conclude` uses `willSendLte(story.draw)` on the treasury (covers both the paying and non-paying outcomes without permitting more), and non-fund-moving calls (`propose`, `vote`, `veto`) use `PostConditionMode.Deny` with an empty condition list so any unexpected asset movement fails the tx. `legion_faucet` is the one `Allow`-mode call, and the comment explaining why (mint has no sender to write a condition against) is correct.
- Local pre-flighting of contract preconditions (`proposeBlockers`, phase checks in vote/veto/conclude) so a doomed call fails before burning gas on a revert, with `explainError` turning `uXXX` reverts into sentences via the `LEGION_ERRORS` map.
- The parent/child inscription ownership re-check between commit and reveal (`legion.tools.ts` inscribe_reveal) — the parent can move in the gap between broadcasts, and re-verifying instead of trusting the earlier lookup avoids a broadcast that fails after the commit fee is already spent.
- `predictOutcome`'s settled-vs-forecast branch (`legion.service.ts:781-792`) is a subtle correctness point handled right: a passed proposal is definitionally past its conclude window, so running the forecast branch on a settled story would mislabel it "expired." Good catch, and the comment explains why.

**Code quality notes:**
- `[nit]` `legion_sponsor` (`src/tools/legion.tools.ts` ~line 1710-1715) calls `getLegionAccount()` twice — once into `account`, then again inline for the balance check (`sbtcBalance((await getLegionAccount()).address)`). Every other tool reuses the first `account.address`; this one could too:
```suggestion
          const [token, minSponsor, balance] = await Promise.all([
            getLegionToken(),
            readTreasury("get-min-sponsor"),
            sbtcBalance(account.address),
          ]);
```
  Cheap operation either way (no extra chain call, just a redundant key derivation), not blocking.

**Operational context:** we track a related mainnet-cut decision from Quasar Garuda's News Legion ask (aibtcdev/legions#12) — this PR stays entirely on testnet (`news-gov-v5-testnet`), so it doesn't touch that open question about mainnet sBTC funding. Good sequencing to land the tooling before that's resolved.

No blocking issues across functionality, security, performance, or big-picture fit. Approving.

PR #647, opened 2026-08-05
feat(legion): require explicit consent before spending mainnet BTC

biwasxyz:
The legion's governance is Stacks testnet, so an agent working through these tools has no reason to expect real money to move. Inscription is the one step that does.

Only 2 of the 13 tools can reach mainnet at all:

- The **11 governance tools** use `LEGION_NETWORK`, pinned from the contract address. They cannot touch mainnet — unchanged here.
- **`legion_inscribe_story` / `legion_inscribe_reveal`** use the global `NETWORK` and spend real BTC.

Inscription cannot be moved off mainnet: ordinals.com indexes mainnet only (`/status` reports `chain: mainnet`; `signet.`/`testnet.` don't resolve), so a testnet inscription produces a link no voter can open. Blocking it outright would break the publish path. So the cost is quoted and consent taken instead.

**`legion_inscribe_story`** now refuses on mainnet without `confirmMainnetSpend: true`:

```
legion_inscribe_story { title, body }
  → Error: This would spend 4750 sats of REAL BITCOIN on mainnet — 1400 commit
    fee plus 3350 locked for the reveal. The legion's governance is Stacks
    testnet, but the inscription is mainnet Bitcoin because ordinals.com indexes
    mainnet only. Nothing was signed. Re-call with confirmMainnetSpend: true to
    proceed, or dryRun: true for the full breakdown.
```

The gate sits **after** the commit is built, so the refusal carries exact sats, and **before** anything is signed — build (1484) → dryRun (1547) → gate (1564) → sign (1578) → broadcast (1582).

**The reveal is deliberately not gated.** Those sats are already committed at that point; refusing there would strand them rather than save them. Called out in the tool description so it reads as a decision rather than an oversight.

## Testing

`tsc` and build pass. Verified through the server with `NETWORK=mainnet`: the new `confirmMainnetSpend` param is exposed on the commit and absent from the reveal, and `legion_status` still reports `testnet`.

The refusal itself is not exercised at runtime — it sits after the commit build, so reaching it needs a funded mainnet wallet. Verified by ordering instead.

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -663,6 +663,17 @@ must be inscribed **one at a time**. Ownership is checked before the commit is
 broadcast and re-checked before the reveal, since the parent can move in
 between. Use `estimate_child_inscription_fee` for the extra input/output cost.
 
+**Only two of these tools can reach mainnet.** The eleven governance tools are
+pinned to the legion's chain by contract address and cannot touch mainnet at
+all. `legion_inscribe_story` and `legion_inscribe_reveal` follow the global
+`NETWORK` and spend real BTC — inscription cannot be moved off mainnet, because
+ordinals.com indexes mainnet only.
+
+So the commit asks for consent: on mainnet it prices the inscription, refuses,
+and reports the exact sats unless `confirmMainnetSpend: true` is passed. The
+**reveal is deliberately not gated** — those sats are already committed, and
+refusing there would strand them rather than save them.
+
 **Inscription pre-flight.** Bitcoin failures cost real sats and 10–60 min per
 confirmation, so `legion_inscribe_story` refuses locally before it signs:
 
@@ -671,6 +682,7 @@ confirmation, so `legion_inscribe_story` refuses locally before it signs:
 | content ≤ 390,000 bytes | the body rides in the reveal witness — oversized commits fine, then can never be revealed |
 | fee estimate is finite and positive | the builders only reject `<= 0`; `NaN` slips past into every size calculation |
 | `NETWORK === "mainnet"` unless `allowNonMainnet` | ordinals.com indexes mainnet only, so the link would 404 for every voter |
+| `confirmMainnetSpend: true` on mainnet | the only step that moves real money; quoted and consented to, never implicit |
 | parent is held by this wallet | the reveal must spend the parent's UTXO |
 | funding covers reveal amount + commit fee | from the builder, with exact numbers |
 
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Adds an explicit consent gate before `legion_inscribe_story` spends real BTC on mainnet — good addition, and it lines up with how we handle irreversible-spend decisions on our own side (we escalate mainnet sBTC commitments to a human rather than auto-approving them).

**What works well:**
- The gate sits in the right place: *after* `commitResult` is built (so the refusal can quote exact sats) and *before* `signBtcTransaction` (so nothing is ever signed without consent). That ordering is easy to get wrong — it'd be tempting to gate earlier for a "cheap" fail-fast, but that would mean the error message can't state a real cost.
- `dryRun` still short-circuits before the new gate, so a caller can preview cost/probes without passing `confirmMainnetSpend` — no annoying double-flag requirement (`dryRun: true` *and* `confirmMainnetSpend: true`) just to look at the numbers.
- Correctly leaves `legion_inscribe_reveal` ungated. Gating the reveal would strand already-committed sats rather than protect anything — the doc comment explaining that (`docs/TOOLS.md` and the inline comment) makes the asymmetry obvious to future readers instead of looking like an inconsistency.
- The eleven governance tools are untouched, which is correct — they're pinned by contract address and structurally can't reach mainnet, so gating them would just be noise.
- `docs/TOOLS.md` and the tool description string are updated in the same PR, and the pre-flight table gets a new row — docs won't drift from behavior here.

**[question]** `confirmMainnetSpend` is a plain boolean the caller passes back, not tied to the specific quoted amount from a prior `dryRun`/probe call. An agent (or the LLM operating it) could pass `confirmMainnetSpend: true` up front on the very first call without ever seeing the cost, skipping the "quote then confirm" flow the design intends — functionally that's an "I consent to whatever it costs" flag rather than "I consent to *this* cost." Is that acceptable for this tool's threat model (a caller who already knows roughly what inscriptions cost), or would it be worth tightening later (e.g. requiring the confirmation to echo back the quoted `totalSats`)? Not blocking — mirrors how `allowNonMainnet` already works, so it's at least consistent with existing style.

**Code quality notes:** nothing to flag — the diff is small, the new branch reuses `commitResult`/`createErrorResponse` that already existed, and the comments explain *why* (testnet governance vs. mainnet inscription, why reveal isn't gated) rather than restating what the code does.

*Operational context:* we've hit this exact class of problem from the caller side — our own news-legion review (`aibtcdev/legions#12`) explicitly withheld any mainnet sBTC commitment pending human sign-off, specifically because nothing on the tool side forced that pause. This PR is the missing enforcement on the other end of that gap, so it directly reduces the risk we flagged there.

Issue #649, opened 2026-08-05
legion_status / legion_list_stories / legion_get_story / legion_my_position pinned to retired news-gov-v5-testnet, not live v6


sonic-mast:
## What's wrong

aibtcdev/legions PR #13 ("news legion v6: Clarity 5, no veto, 8 pieces/day drain ceiling") merged 2026-08-05T11:20Z and deployed a new governance/treasury pair at a different deployer address:

- v6 (live): `ST2VN1G6EBXPMMAJKCSY1HR50YQCVFSK68KKP9SKW.news-gov-v6-testnet` / `...news-treasury-v6`
- v5 (retired): `STGX5YP51NKM69ZMP6DVB6GAJAANCG5WB3718KD9.news-gov-v5-testnet` / `...news-treasury-v5`

`GET https://aibtc.news/api/state` already reflects this correctly — its `legions[]` array has the v6 entry with `"live": true` and the v5 entry with `"live": false`.

But the native MCP tools (`legion_status`, and by extension `legion_list_stories`, `legion_get_story`, `legion_my_position`) are still hardcoded/wired to the retired v5 contract address. Confirmed live just now:

```json
{
  "contracts": {
    "governance": "STGX5YP51NKM69ZMP6DVB6GAJAANCG5WB3718KD9.news-gov-v5-testnet",
    "treasury": "STGX5YP51NKM69ZMP6DVB6GAJAANCG5WB3718KD9.news-treasury-v5"
  },
  "pool": {"balanceSats": 30039847},
  "governance": {"totalWeight": 30000000, "lastProposalId": 5}
}
```

Meanwhile `/api/state` shows v6 already has 6 real proposals, active voting (one at 29M yes / 10M weight), and a 49,000,000 sat pool — none of which any `legion_*` caller can currently see.

## Impact

Any agent relying on `legion_status`/`legion_list_stories`/`legion_get_story`/`legion_my_position` for decision-relevant reads (pool balance, live proposals, governance params, own weight/position) is silently looking at a dead contract. New v6 params also differ materially from v5 (`votingQuorum` 15→10, `minParticipants` 2→1, veto removed entirely) — an agent using v5 params to reason about v6 proposals would draw wrong conclusions about quorum/veto risk.

## Suggested fix

[...]

PR #651, opened 2026-08-06
feat(legion): support v6, and make eras a first-class concept

biwasxyz:
v6 is deployed (`ST2VN1G6EBXPMMAJKCSY1HR50YQCVFSK68KKP9SKW.news-gov-v6-testnet`) and news-legion's `v6-contract` branch already treats it as primary with v5 retained. The published tools were pinned to v5 and would have broken against it.

## What v6 changes

| | v5 | v6 |
|---|---|---|
| `veto` | public fn | **removed entirely** |
| `vote` | `(proposalId, support)` | `(proposalId, support, rationale)` — non-empty, `u440` |
| `get-params` | 12 fields | 10 — no `vetoQuorum` / `vetoWindow` |
| `get-story` | `vetoWeight` | `concluded` |
| quorum / participants | 15% / 2 | 10% / 1 |
| new views | — | `vote-power (proposalId who)` |
| deployer | `STGX5…` | `ST2VN1…` (different) |
| proposal ids | 1..5 | restart at 1 |

Concretely, the 1.67.0 tools against v6: `legion_veto` calls a function that no longer exists; `legion_status` and `legion_get_story` read `vetoQuorum`/`vetoWindow` as `NaN`; `predictOutcome` compares against `NaN`; `getVetoRecord` calls a missing view.

## Approach

`LEGION_ERAS` lists deployments live-first. Writes go to the live era — proposing into a retired contract would govern a deployment nothing watches. Reads take an optional `era`, because a bare proposal id means different stories in v5 and v6. Treasury is derived per era from the gov deployer, since the two eras have different ones.

**Per-era differences are read from the contract interface, not inferred from the version number.** `getCapabilities` asks the deployed contract whether a `veto` function exists and how many arguments `vote` takes. A later era that changes again needs one line in `LEGION_ERAS` and no other code.

Veto params are `?: number` rather than defaulted — an era without veto skips the branch instead of comparing against zero and silently passing.

## Testing

Exercised against both live eras through the MCP server:

```
LIVE ERA: v6, supportsVeto:false, voteNeedsRationale:true
  params: quorum 10, participants 1, no veto keys
ERA 5:    supportsVeto:true, vetoQuorum 15
v6 stories: #1/expired (total 1)   v5 stories: #5,#4,#3 passed (total 5)
v6 story 1: vetoWeight undefined, outcome expired/not-concluded
v5 story 5: vetoWeight 0, outcome passed/paid

legion_veto on v6  → "Legion v6 has no veto … Vote no with legion_vote instead."
legion_vote no rationale → "records a rationale with every vote and rejects an empty one (u440)"
era: 99 → "No legion era v99. Known eras: v6 (live), v5"
```

`tsc` and build pass. Write paths still never broadcast — needs testnet sBTC and a funded key.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -219,7 +219,7 @@ The allowlist is re-enforced at `tools/call` time, so the model can't reach a to
 | Competition | `competition_submit_trade/status/list_trades/allowlist` | Mainnet only; requires identity registration; Bitflow swaps only score today |
 | Pillar | `pillar_connect/disconnect/status/send/fund/supply/boost/unwind/auto_compound/position/create_wallet/add_admin/invite` | Browser handoff for passkey signing |
 | AIBTC News | `news_list_signals/front_page/leaderboard/check_status/list_beats` (read) + `news_file_signal/claim_beat` (BIP-322 auth) | bc1q addresses only |
-| News Legion | `legion_status/list_stories/get_story/my_position` (read) + `legion_contribute/sponsor/propose_story/vote/veto/conclude/faucet` + `legion_inscribe_story/inscribe_reveal` | news-gov-v5 on Stacks **testnet**, pinned by contract address — never follows global `NETWORK`. Inscription is the exception: real BTC on whatever `NETWORK` names |
+| News Legion | `legion_status/list_stories/get_story/my_position` (read) + `legion_contribute/sponsor/propose_story/vote/veto/conclude/faucet` + `legion_inscribe_story/inscribe_reveal` | Multi-era: **v6 live** (writes), v5 readable via `era: 5`. Stacks **testnet**, pinned by contract address — never follows global `NETWORK`. v6 dropped veto and requires a vote `rationale`; per-era features are read from the contract interface. Inscription is the exception: real BTC on whatever `NETWORK` names, and needs `confirmMainnetSpend` |
 | Inbox | `send_inbox_message_direct` | Mainnet only; non-sponsored sBTC transfer, sender pays STX gas. `send_inbox_message` (sponsored relay path) is **deprecated** — it no longer sends and just redirects here (relay queue could wedge, #540/#592) |
 
 ## Agent Behavior Guidelines
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,9 +96,9 @@ aibtc-mcp-server MCP Server (src/index.ts)
 - `src/services/scaffold.service.ts` - x402 endpoint project scaffolding for Cloudflare Workers
 - `src/tools/bitcoin.tools.ts` - Bitcoin L1 tools (balance, fees, UTXOs, transfer)
 - `src/tools/news.tools.ts` - AIBTC News tools (signals, beats, briefs, BIP-322 auth; signal filing is free, x402 payment kept as fallback)
-- `src/tools/legion.tools.ts` - AIBTC News Legion governance (inscribe → propose → vote/veto → conclude, contribute, sponsor)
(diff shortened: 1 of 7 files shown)


arc0btc (approved):
Adds multi-era support for News Legion — v6 goes live, v5 stays readable for history. This is the fix for the exact blocker we flagged in `aibtc-mcp-server#649`: the previously-published `legion_*` tools were hardcoded to `news-gov-v5-testnet`, which silently reads a dead pool now that `news-gov-v6-testnet` is live (veto removed, `vote` requires a `rationale`, different deployer). Confirms our earlier assumption that the on-chain read path (`readGov`) was sound in isolation — the bug was always the pinned contract constant, not the read logic itself.

**What works well:**
- `getCapabilities()` (`src/services/legion.service.ts:345`) reads `veto`/`vote` arity/`vote-power` straight off the deployed contract interface instead of branching on the version number. That's the right call given the PR's own framing — "a later era that changes again needs one line in `LEGION_ERAS` and no other code" — and it's actually true given how the diff is structured.
- Making `vetoQuorum`/`vetoWindow`/`vetoWeight` genuinely `undefined` rather than defaulting to `0` (`legion.service.ts:316-319`, `:465-468`) is the correct fix for the "silent NaN" failure mode described in the PR body. `predictOutcome`'s `vetoSupported` gate (`legion.service.ts:614-617`) and `buildTimeline`'s `vetoEnd` handling (`:596-606`) both correctly skip the veto branch instead of comparing against a fabricated zero.
- Every cache (`paramsCache`, `timingModeCache`, `tokenCache`, `capabilitiesCache`) is keyed by era (`legion.service.ts:334-337`) — avoids the cross-era cache-poisoning bug that a naive single-value cache would have hit the moment `getParams()` was called for both v5 and v6 in the same process.
- Writes are hardcoded to `LIVE_ERA` (`legion.tools.ts:687-689`) rather than accepting an `era` override, so there's no way to accidentally propose/vote into a retired contract that nothing watches. Good boundary — read tools get the flexible `era` param, write tools don't.
- The network-consistency guard at module load (`config/legion.ts:210-216`) — throwing if any era's contract isn't on the same chain as the live one — is cheap insurance against a future config mistake wiring in a mainnet id by accident.
- Docs (`CLAUDE.md`, `docs/TOOLS.md`) were updated in the same PR and match the code, including the new era table and the `legion_veto` v5-only caveat.

**Operational context:** this unblocks Arc's own proposer/voter/veto automation plan, which has been sitting behind #649 since the v6 fork (`news-legion-mainnet-sbtc-contribution` thread). Once merged we can re-evaluate wiring role automation — though the mainnet sBTC contribution + cut-timing question is still a separate, unrelated whoabuddy sign-off.

**[question]** `legion_vote` and `legion_veto` call `getCapabilities()` with no era argument (`legion.tools.ts:1045`, `:1094`), relying on the default parameter resolving to `LIVE_ERA`. That's correct today since writes only ever target the live era, but it reads slightly implicit next to the read tools where `era` is threaded explicitly end-to-end. Not asking for a change — just confirming that's intentional rather than an oversight, since a future refactor that adds an `era` param to a write tool without noticing this default could silently check capabilities against the wrong contract.

**[nit]** `versionOf`'s regex `/\bv(\d+)\b/` matches against the full contract id (deployer address + name). Deployer addresses are uppercase-only c32, so there's no collision today, but a lowercase-address chain or a differently-cased future network would be worth a passing thought if this ever gets reused outside Stacks testnet/mainnet c32 addresses.

**Code quality notes:** No reuse or over-engineering concerns — the abstraction level (interface-driven capability detection over version-number branching) matches the stated goal and the diff doesn't introduce speculative generality beyond what two live eras already require.

Testing section in the PR description (exercised against both live eras, `tsc`/build pass) covers what I'd have asked for. Approving.

PR #655, opened 2026-08-13
feat: add TaskMarket work-marketplace tools (search/get/submissions/stats + gated create)

DinhTruong98:
## Summary

Adds a [TaskMarket](https://taskmarket.dev) integration to the MCP server. TaskMarket is an onchain agent work marketplace on Base: requesters escrow USDC, workers submit deliverables, and the requester reviews and accepts a winner. These tools let an agent browse that market and, with explicit operator authorization, post work to it.

## Why

When a request is better delegated to a competitive worker market than burned on local inference, the agent can now search TaskMarket, inspect a task, and create one with a real spending gate — instead of guessing or linking to a homepage. Read tools are fully public and anonymous; the one write tool escrows real USDC and is gated hard.

## New tools

**Read-only (public API, anonymous, no wallet, no spend):**
- `taskmarket_search` — list/search open tasks (status, phase, mode, tags, reward range)
- `taskmarket_get` — full task detail + live status + link
- `taskmarket_submissions` — list submissions for **human review** (never auto-accepts or auto-rejects)
- `taskmarket_stats` — public agent/market statistics (reputation check)

**Gated write (escrows real USDC):**
- `taskmarket_preview_create` — show the exact plan (description, reward, deadline, deliverables, Base network, max spend) and prove the confirmation gate works **before any money can move**
- `taskmarket_create` — create + fund via the first-party TaskMarket CLI (USDC escrow on Base via x402) after explicit `confirm="APPROVE"` + a `maxSpendUsdc` cap; returns task ID, link, and live status

## Safety invariants

- Create **never** runs without an explicit confirmation token supplied fresh by the operator — never inferred from prompt content.
- The reward must be `<=` the caller's `maxSpendUsdc` cap; otherwise the create refuses with "No funds were moved".
- Network pinned to Base (`eip155:8453`), USDC `0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913`.
- No blind retry of unknown-settlement payments: a failed or task-id-less create is surfaced (check `taskmarket inbox`), never silently re-submitted.
- No private keys, seeds, tokens, or cookies are ever requested, stored, or logged.

## Verification

- `npm run build` — clean
- `npm test` — 583 passing (was 552 before; +31 new hermetic tests for the flow + safeguards in `tests/services/taskmarket.test.ts` and `tests/tools/taskmarket.test.ts`)
- `node scripts/taskmarket-live-check.mjs` — boots the real server over stdio and calls all tools against the live TaskMarket public API; both refusal paths (no confirm, reward>cap) verified live with no funds moved

## Files

- `src/services/taskmarket.service.ts` — read-only API client + `prepareCreatePlan` gate
- `src/tools/taskmarket.tools.ts` — MCP tool registrations
- `tests/services/taskmarket.test.ts`, `tests/tools/taskmarket.test.ts` — hermetic tests
- `scripts/taskmarket-live-check.mjs` — live smoke script
- Docs: `README.md`, `docs/TOOLS.md`, `CLAUDE.md`, `skill/SKILL.md`, `skill/references/taskmarket.md`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -204,6 +204,7 @@ The allowlist is re-enforced at `tools/call` time, so the model can't reach a to
 | Discovery | `list_x402_endpoints` | Start here to find x402 actions |
 | Earning | `earning_opportunities` | Static "how to put your assets to work" menu; surface after `identity_register` |
 | Bounties | `bounty_list/get/submissions` (read) + `bounty_submit/create/accept/paid/cancel` (BIP-322 auth) + `bounty_my_posted/my_submissions` | aibtc.com sBTC bounty board; most direct way to earn. Submit needs Registered (L1+); create needs Genesis (L2+); bc1q/P2WPKH signing; accepted submission paid in sBTC |
+| TaskMarket | `taskmarket_search/get/submissions/stats` (read, public) + `taskmarket_preview_create` + `taskmarket_create` (gated) | Onchain agent work marketplace on Base (USDC escrow). Read tools are anonymous and free. `taskmarket_create` requires explicit `confirm="APPROVE"` + `maxSpendUsdc` cap, pins Base network, refuses otherwise, and never blindly retries unknown-settlement payments. See `docs/TOOLS.md` |
 | Wallet & balance | `get_wallet_info`, `get_stx_balance` | |
 | Wallet mgmt | `wallet_create/import/unlock/lock/list/switch/delete/export/status` | On mainnet, create/import also derive a Spark Lightning wallet from the same mnemonic |
 | Bitcoin L1 | `get_btc_balance/fees/utxos`, `transfer_btc` | mempool.space; P2WPKH; sats |
@@ -232,6 +233,7 @@ When a user asks for something:
 4. **For any x402 URL** → Use `execute_x402_endpoint` with full `url` parameter - works with ANY x402-compatible endpoint
 5. **For Pillar smart wallet actions** → Use `pillar_connect` first, then `pillar_send`, `pillar_fund`, `pillar_boost`, etc.
 6. **For aibtc.news actions** → Use `news_list_beats` to discover beats, then `news_file_signal` to file (filing is free; falls back to x402 payment if the endpoint requires it)
+7. **For TaskMarket work** → Use `taskmarket_search` to browse open tasks, `taskmarket_get`/`taskmarket_submissions` to inspect a task and its submissions for review. To create a task, first call `taskmarket_preview_create` (shows the exact plan + confirms the gate), then `taskmarket_create` with the same details plus `confirm="APPROVE"` and a `maxSpendUsdc` cap — it refuses otherwise (no funds moved)
 7. **For News Legion governance** → Start with `legion_status`, then `legion_list_stories`. To publish: `legion_inscribe_story` → `legion_inscribe_reveal` → `legion_propose_story`. To judge: `legion_get_story` (open its `contentUrl` and read the piece) → `legion_vote` / `legion_veto` → `legion_conclude`
 8. **For unknown actions** → Ask user for the x402 endpoint URL or check if it's a direct blockchain action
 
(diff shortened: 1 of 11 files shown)


arc0btc (approved):
Adds a TaskMarket (onchain agent work marketplace on Base) integration — four public read-only tools plus a hard-gated create path. This is exactly the shape we'd want for a payment-adjacent MCP tool: cheap discovery is fully anonymous, and the only tool that moves money forces an explicit `confirm="APPROVE"` + a `maxSpendUsdc` cap the reward has to fit under, with `prepareCreatePlan` centralizing all the validation so both `taskmarket_preview_create` and `taskmarket_create` share one gate.

**What works well:**
- `execFile` (not `exec`/shell) with an args array in `src/tools/taskmarket.tools.ts` — no shell interpolation of `description`/`tags`, so arbitrary task descriptions can't break out into shell metacharacters. This is the right call given `description` is free text.
- `prepareCreatePlan` is pure and side-effect-free (`src/services/taskmarket.service.ts:255`) — it only returns a plan or a refusal, never touches the network. `taskmarket_preview_create` and `taskmarket_create` both call through it, so the confirm/cap logic can't drift between the two tools.
- The confirm-token check is an exact string match (`opts.confirm.trim() !== "APPROVE"`), not a truthy/fuzzy check, and 31 hermetic tests exercise the gate directly (missing token, wrong token, reward > cap, zero reward, short description, bad duration, bad visibility, base-unit cap variant).
- `taskmarket_create` treats a failed/parseless CLI response as unknown-settlement and explicitly does not retry, telling the caller to check `taskmarket inbox` instead — matches our own no-blind-retry-of-unknown-settlement pattern for x402/payment flows.
- `TASK_ID_SCHEMA` regex-validates task IDs (`0x` + 64 hex) before they reach the API/CLI.

**[question] Confirm token is just another LLM-supplied tool argument** (`src/tools/taskmarket.tools.ts:248,311`)
The docstring says "supplied fresh by the operator... never inferred from prompt content," but structurally `confirm` is a normal Zod string field in the tool's `inputSchema` — whatever calls the MCP tool (which may itself be an LLM agent) can just pass `confirm: "APPROVE"` in the same tool call. There's nothing in this file that ties the token to an actual human-in-the-loop step (no out-of-band code, no separate approval channel). Is the human-authorization boundary enforced by the *caller* (e.g., a host app that intercepts this specific field before it reaches the LLM), or does this PR rely on the agent itself behaving well? If it's the former, worth a one-line comment here since it's not obvious from this file alone; if it's the latter, this is the same "gate is prompt-reachable" shape we've flagged before in our own gated-write tools, and worth knowing about even if not blocking for this PR.

**[suggestion] No timeout on the read-only `fetch` calls** (`src/services/taskmarket.service.ts:123`)
`apiGet` takes an optional `AbortSignal` but none of the four read tools pass one in — a slow/hung TaskMarket API response has no bound and will hold the MCP call open indefinitely. `taskmarket_create`'s CLI invocation gets a `CLI_TIMEOUT_MS` (120s); the read path deserves the same discipline, e.g. wiring `AbortSignal.timeout(15_000)` in each tool handler.

**[suggestion] `env: { ...process.env }` is redundant** (`src/tools/taskmarket.tools.ts:351`)
`execFile` inherits the parent process's environment by default when no `env` option is passed at all, so spreading `process.env` explicitly doesn't change behavior — it just makes it easy to later add a key here without noticing the whole environment (including any unrelated secrets in `process.env`) is already being forwarded. Not a bug today, but if the CLI ever needs a scoped subset of env vars, this line is where that would need to change, and its current form doesn't signal that.

**Code quality notes:**
- `usdcToBaseUnits`/`baseUnitsToUsdc` are clean, well-tested round-trip helpers with explicit precision limits (6 decimals) — no complaints there.
- `taskmarket_preview_create` and `taskmarket_create` duplicate the ~10-line `cliArgs` construction verbatim. Small enough that extracting a `buildCreateCliArgs(plan)` helper is optional, but it would remove the risk of the two argument lists drifting if a new field is added later.

**Operational context:** we don't currently hold a TaskMarket-CLI-compatible wallet, so we can't live-test the write path from our side, but the read tools' anonymous/no-wallet design matches how we'd want to probe a marketplace before deciding whether to delegate anything to it.

No blocking issues — the safety invariants described in the PR body are actually implemented and tested, not just asserted in prose. Approving.

PR #656, opened 2026-08-17
feat(legion): move the News Legion tools to mainnet

biwasxyz:
The News Legion is live on Stacks mainnet with real sBTC:

- `SP5Y3W3F78NKFH4HYFNDQMJC484VZWKDH35ZR2M9.aibtc-news-gov`
- `SP5Y3W3F78NKFH4HYFNDQMJC484VZWKDH35ZR2M9.aibtc-news-treasury`

Supersedes #651, whose branch was renamed (that closed it). This branch still
contains its commit, so this PR is the full diff against `main`.

## What the mainnet contract changed

It is a rewrite, not a redeploy. Against the testnet v6 build:

| | testnet v6 | mainnet |
|---|---|---|
| clock | `TEST-STACKS-BLOCKS` | `PROD-BURN` — burn blocks |
| quorum | 10% of `eligibleSnapshot` | **removed entirely** |
| new gate | — | `yesMultiple: 20` — yes weight must cover 20x the payout, else `yes-short` |
| new gate | — | `membersToActivate: 21` — no proposals until 21 members (`u441`) |
| params | `minWeight`, `minContribution`, `drawBps`, `votingDelay`, `proposeInterval`, `minParticipants` | `minWeightToAct`, `minJoinSats`, `payoutBps`, `voteDelay`, `globalProposeInterval`, `minVoters` |
| story tuple | `bond`, `draw`, `eligibleSnapshot` | `lockedWeight`, `payout`, `totalWeightAtOpen` (now total, incl. proposer) |
| reads | `quote-draw`, `locked-of`, `bond-unlock-at` | `quote-payout`, `get-locked-weight`, `get-locked-until` |
| new reads | — | `get-member-count`, `is-activated` |
| names | `news-gov-v6-testnet` | `aibtc-news-gov` — **no version suffix** |

The name change alone broke `versionOf()` and `treasuryFor()`, which parsed the
version out of the contract name.

## Approach

Testnet v5/v6 are dropped rather than kept readable. Keeping them would mean a
per-era network on every read path, and they sit on a chain that has already
been reset once. `LEGION_ERAS` is a single mainnet entry at version 1, declared
rather than parsed. Adding a future generation is one entry; an unknown `era`
errors instead of silently answering about the live one.

`getCapabilities` and `getVetoRecord` are removed. They existed to let two eras
differ; with one era of known shape they read the chain to answer a constant.

`predictOutcome` follows the contract's decision order exactly — voters ->
threshold -> yes weight -> pool. The yes-weight gate is the subtle one: a piece
can be approved 100% and still fail, because the yes side must stake 20x the
sats it releases.

`legion_veto` and `legion_faucet` are deleted, not stubbed — the contract has no
`veto` function and real sBTC has no faucet.

## Safety: sBTC spends are now metered

`docs/TOOLS.md` previously stated legion spends were exempt from the
`SPEND_LIMIT_*` rail "because these are testnet sBTC". That is no longer true.
`legion_contribute` and `legion_sponsor` move real, non-refundable sBTC, so both
now `check`/`record` against the `sats` ledger before signing — they take the
amount as a direct parameter, which is what the rail needs.

Consequence worth knowing: the treasury's minimum sponsorship is 100,000 sats
and the default session cap is 50,000, so a first `legion_sponsor` blocks by
design until `SPEND_LIMIT_SESSION_SATS` is raised.

## Testing

`tsc` clean, build clean, 576 tests pass (24 new).

Every read verified against the live contracts through the MCP server:

```
network mainnet, era v1, timingMode PROD-BURN, height 962865
params: threshold 66, minVoters 1, membersToActivate 21, yesMultiple 20,
        minWeightToAct/minJoinSats 10000, payoutBps 5,
        voteDelay 2, voteWindow 30, concludeWindow 12, globalProposeInterval 18
membership: 1/21, activated false
treasury wired to .aibtc-news-gov, token SM3VDXK3....sbtc-token
era: 6 -> "No legion era v6. Known eras: v1 (live)"
```

New tests cover each conclude branch (`no-voters`, `voted-down`, `yes-short`,
`pool-short`, `paid`, `expired`), that threshold is tested before yes weight
when both fail, the settled-story path, and the shared-deployer invariant.

Write paths are not exercised end-to-end: the legion sits at 1/21 members, so
`propose-story` reverts with `u441` until 20 more join.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -219,7 +219,7 @@ The allowlist is re-enforced at `tools/call` time, so the model can't reach a to
 | Competition | `competition_submit_trade/status/list_trades/allowlist` | Mainnet only; requires identity registration; Bitflow swaps only score today |
 | Pillar | `pillar_connect/disconnect/status/send/fund/supply/boost/unwind/auto_compound/position/create_wallet/add_admin/invite` | Browser handoff for passkey signing |
 | AIBTC News | `news_list_signals/front_page/leaderboard/check_status/list_beats` (read) + `news_file_signal/claim_beat` (BIP-322 auth) | bc1q addresses only |
-| News Legion | `legion_status/list_stories/get_story/my_position` (read) + `legion_contribute/sponsor/propose_story/vote/veto/conclude/faucet` + `legion_inscribe_story/inscribe_reveal` | news-gov-v5 on Stacks **testnet**, pinned by contract address — never follows global `NETWORK`. Inscription is the exception: real BTC on whatever `NETWORK` names |
+| News Legion | `legion_status/list_stories/get_story/my_position` (read) + `legion_contribute/sponsor/propose_story/vote/veto/conclude/faucet` + `legion_inscribe_story/inscribe_reveal` | Multi-era: **v6 live** (writes), v5 readable via `era: 5`. Stacks **testnet**, pinned by contract address — never follows global `NETWORK`. v6 dropped veto and requires a vote `rationale`; per-era features are read from the contract interface. Inscription is the exception: real BTC on whatever `NETWORK` names, and needs `confirmMainnetSpend` |
 | Inbox | `send_inbox_message_direct` | Mainnet only; non-sponsored sBTC transfer, sender pays STX gas. `send_inbox_message` (sponsored relay path) is **deprecated** — it no longer sends and just redirects here (relay queue could wedge, #540/#592) |
 
 ## Agent Behavior Guidelines
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,9 +96,9 @@ aibtc-mcp-server MCP Server (src/index.ts)
 - `src/services/scaffold.service.ts` - x402 endpoint project scaffolding for Cloudflare Workers
 - `src/tools/bitcoin.tools.ts` - Bitcoin L1 tools (balance, fees, UTXOs, transfer)
 - `src/tools/news.tools.ts` - AIBTC News tools (signals, beats, briefs, BIP-322 auth; signal filing is free, x402 payment kept as fallback)
-- `src/tools/legion.tools.ts` - AIBTC News Legion governance (inscribe → propose → vote/veto → conclude, contribute, sponsor)
+- `src/tools/legion.tools.ts` - AIBTC News Legion governance (inscribe → propose → vote → conclude, contribute, sponsor)
 - `src/services/legion.service.ts` - Legion chain reads, network-pinned account, phase/outcome derivation
-- `src/config/legion.ts` - Legion contract ids (constants), network derivation, contract error codes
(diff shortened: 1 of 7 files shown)


arc0btc (approved):
Moves the News Legion tools from testnet v5/v6 to the live mainnet contracts (`aibtc-news-gov` / `aibtc-news-treasury`), and it's a rewrite, not a rename — quorum and veto are gone, replaced by a member-activation gate (21 members) and a yes-weight-vs-payout multiplier (20x). This is squarely in our own operational path (we hold seat 1/21 on this exact legion), so I read it closely.

**What works well:**
- `LEGION_ERAS` (`src/config/legion.ts`) is the right shape for this problem — a declared array with a `live` flag and a module-load-time invariant check (same chain, gov/treasury share a deployer) beats parsing versions out of contract names, especially now that mainnet dropped the version suffix entirely.
- `resolveEra()` erroring on an unknown version instead of silently falling back to the live era is exactly the right call — a wrong era would answer about a different proposal with the same id.
- `predictOutcome`'s gate order (voters → threshold → yes-weight → pool) matches the contract's own decision order per the PR description, and the test at legion.test.ts covers that ordering explicitly ("checks threshold before yes weight, matching the contract's order") rather than just asserting the end state.
- Real money handling is careful: `legion_contribute`/`legion_sponsor` now `check()` the spend-limiter before `callContract` and `record()` only after it returns — check-before-sign, record-after-confirm-broadcast is the correct order for a non-refundable spend.
- Removing `legion_veto`/`legion_faucet` entirely (not stubbing) matches "the contract has no veto function, real sBTC has no faucet" — I grepped the new tools file for leftover veto/faucet references and found none.
- Good test coverage on the pure logic (`predictOutcome`, `proposeBlockers`, `buildTimeline`, era resolution) — 24 new tests per the PR description, and they exercise the interesting edge cases (yes-short vs voted-down, settled-vs-forecast, turnout not failing a story alone).

**[question]** `legion_vote`'s rationale blank-check (`rationale.trim().length === 0`) runs after the `Promise.all` that fetches story/phase/weight/etc. Since the zod schema already enforces `min(1)`, this can only fire on an all-whitespace string — not a bug, just means a few wasted RPC reads on that specific input. Not worth blocking on, just flagging since I noticed it.

**[nit]** `docs/TOOLS.md`'s "Things that bite" section is comprehensive and calls out the activation gate and the yes-multiple gate up front — appreciated, since both are the kind of thing that's easy to miss reading the tool descriptions alone.

**Operational note:** we're currently the only member of this legion (1/21) and are withholding further mainnet sBTC contribution pending human sign-off (irreversible-funds rule) — so `legion_status.membership.activated` will read `false` for us for a while yet. The PR's framing of that gate (`u441`, "contributing more sats yourself does not help") matches what we'd already worked out independently from the raw contract reads.

Solid rewrite, internally consistent across config/service/tools/docs/tests. Approving.

